PR #3, opened 2019-06-16
Fix typo in README.md

MaxDesiatov:


MaxDesiatov:
I don't think the PR failure would be caused by my changes in README.md? 🤔

broadwaylamb:
Thank you @MaxDesiatov!

Indeed the failure is unrelated, it’s just that benchmarking on Travis is a bit unstable.

PR #8, opened 2019-06-22
Get initial pass of working Decode publisher

spadafiva:
@broadwaylamb Okay, here's the PR for `Decode`.  I'm sure that it could use some more tests and I haven't totally understood some of your helper classes in the test suite.  Maybe you can give me some ideas for doing that more cleanly.  

I've copied and pasted some things from `DropWhile`, like this chunk:

```
func receive(subscription: Subscription) {
     _upstreamSubscription = subscription
    subscription.request(.max(1))
     _downstream.receive(subscription: self)
}
```

I'm wondering if we should have an internal base class that implements some of these things so that we don't need to rewrite them again and again, or maybe just some sort of abstracted util so that we don't have code duplication.

Anyway, looking forward to some feedback.


spadafiva:
@broadwaylamb Let me know what else needs to happen before you think this is ready to merge and also if you're working on any of the other publishers/operators/subscribers so that I can tackle something else that you're not working on.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,107 @@
...
+    var customMirror: Mirror { return Mirror(self, children: EmptyCollection()) }
+    
+    init(downstream: Downstream, decoder: Coder) {
+        self._downstream = downstream
+        self._decoder = decoder
+    }
+    
+    func receive(subscription: Subscription) {
+        _upstreamSubscription = subscription
+        subscription.request(.max(1))
```
broadwaylamb: Are you sure we need to request `.max(1)` here? In `DropWhile` I've checked that this is what Apple's Combine does, and it made sense because we wanted to skip some unknown number of elements no matter how much the downstream subscriber had requested.

Here is the test that ensures this behavior for `DropWhile`:
https://github.com/broadwaylamb/OpenCombine/blob/17f9b0a29975106d2f51d5d9b7858c5b1e612019/Tests/OpenCombineTests/PublisherTests/DropWhileTests.swift#L90-L129

I haven't checked this for `Decode`/`Encode`/`Map` yet, so I'd be great if you could write a similar test for the `Decode`/`Encode`/`Map` operator and run it in compatibility mode (and tweak your implementation accordingly). If you feel like it'd require too deep dive that you don't have much time for, it's perfectly fine, I'll do it myself.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,107 @@
...
+        _downstream.receive(subscription: self)
+    }
+    
+    func receive(_ input: Input) -> Subscribers.Demand {
+        do {
+            let value = try _decoder.decode(Downstream.Input.self, from: input)
+            return _downstream.receive(value)
+        } catch {
+            _downstream.receive(completion: .failure(error as! Downstream.Failure))
```
broadwaylamb: It seems like we can avoid this force-cast if we add the `Downstream.Failure == Error` constraint to the class declaration.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    func receive(completion: Subscribers.Completion<Failure>) {
+        switch completion {
+        case .finished:
+            _downstream.receive(completion: .finished)
+        case .failure(let error):
+            // Safe to force unwrap here, since Downstream.Failure can be
+            // either Upstream.Failure or Error
+            _downstream.receive(completion: .failure(error as! Downstream.Failure))
```
broadwaylamb: Same here, can we remove this force-cast?

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Publishers.Encode.swift
+//  
+//
+//  Created by Joseph Spadafora on 6/22/19.
+//
+
+
+extension Publishers {
+    
+    public struct Encode<Upstream, Coder> : Publisher where Upstream : Publisher, Coder : TopLevelEncoder, Upstream.Output : Encodable {
```
broadwaylamb: All that I've mentioned for `Decode` applies here as well, I guess

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -0,0 +1,95 @@
...
+        self._downstream = downstream
+    }
+    
+    var description: String { return "Map" }
+    
+    var customMirror: Mirror { return Mirror(self, children: EmptyCollection()) }
+    
+    func receive(subscription: Subscription) {
+        _upstreamSubscription = subscription
+        subscription.request(.max(1))
```
broadwaylamb: Same, do we need this request?

broadwaylamb (commented):
Okay, sorry for the delay, been busy.

The implementation looks good, thank you!

I've pointed out some things that I think are important to address here. I haven't commented on the code style though, as a better way to enforce it is to use a tool like SwiftLint, which I'm gonna add in a day or two.

>I'm sure that it could use some more tests and I haven't totally understood some of your helper classes in the test suite. Maybe you can give me some ideas for doing that more cleanly.

Yeah, I didn't put much effort into documenting my test helpers, guess I'll do that too.

>I'm wondering if we should have an internal base class that implements some of these things so that we don't need to rewrite them again and again, or maybe just some sort of abstracted until so that we don't have code duplication.

These classes indeed look similar, but they're pretty different implementation-wise, so I think for now we stick with the duplication, and if we really feel the need, we can refactor later.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,44 @@
...
+        ("testDecodeWorks", testDecodeWorks)
+    ]
+    
+    private let jsonEncoder = JSONEncoder()
+    private let jsonDecoder = JSONDecoder()
+    
+    var cancel: Cancellable?
+    
+    func testDecodeWorks() {
+        let promise = XCTestExpectation(description: "decode")
```
broadwaylamb: Since everything should happen synchronously, I think it'd be better not to use the `XCTestExpectation ` API, and go with the synchronous testing instead.

In this case, that means just deleting this line and the `wait(for: [promise], timeout: 1)` line.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,44 @@
...
+@available(macOS 10.15, *)
+final class DecodeTests: XCTestCase {
+    static let allTests = [
+        ("testDecodeWorks", testDecodeWorks)
+    ]
+    
+    private let jsonEncoder = JSONEncoder()
+    private let jsonDecoder = JSONDecoder()
+    
+    var cancel: Cancellable?
```
broadwaylamb: Not sure, why should we make this a stored property? IMO to guarantee the independence of the tests, it's better to move this var into the test method.

broadwaylamb:
@spadafiva it looks like we have a collision :(
@MortyMerr has also been working on the `Map` implementation in #9. It has broader test coverage, so I suggest that we go with @MortyMerr's implementation for `Map`.

I'm still looking forward to merging your `Decode`/`Encode` implementation, though.

I guess we need some sort of task management policy to avoid this in the future.

**UPD**: I've created a [GitHub project](https://github.com/broadwaylamb/OpenCombine/projects/2) so that the contributors could check if something is WIP; updated the README accordingly.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: @broadwaylamb It seems like you asked MortyMerr for a very similar demand test.  Is there any way that we can abstract the demand so that we aren't testing the same thing and writing the same code inside every publisher?

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: That makes sense.  Done.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: Yup

Sources/OpenCombine/Publishers/Publishers.Encode.swift
spadafiva: Fixed

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: Keep in mind that we have to be able to run the tests in exactly the same way against Apple's Combine — that's like 80% why we're writing those tests at all.

We could make a test function generic over some `Publisher`, but we can't use it to test all the publishers. For example, the operators that have map-like semantics (`Map`, `Encode`, `Decode`, `Scan`), i. e. those that preserve the number and periodicity of published events, do not change the demand. For those, we can write a generic test.

However, operators like `DropWhile`, `Filter` etc. may request the different number of elements than their downstream, so we can't naively write a fully generic test that could test all the operators without modification, it is likely better to write ad-hoc tests.

But in my opinion, we should write an ad-hoc test first, and only then make it generic.

broadwaylamb:
@spadafiva I see you've merged #9 into this PR branch. I don't think it's a good idea since #9 will likely undergo changes while in review. You don't want to synchronize with #9 every time there's a change there, it's also a larger commit diff which is harder to review.

Let's make PRs concise and don't check in unrelated changes. Can you please revert commit 0f86564 and remove `Map` from your code since it hasn't yet made it to the master branch?

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: @broadwaylamb Okay, I'm fine with writing an ad-hoc test.  It looks like the tracking subscriber only works for something that maps to Int values.  

Can you clarify this a bit for me: 
>I haven't checked this for Decode/Encode/Map yet, so I'd be great if you could write a similar test for the Decode/Encode/Map operator and run it in compatibility mode (and tweak your implementation accordingly). If you feel like it'd require too deep dive that you don't have much time for, it's perfectly fine, I'll do it myself.

For example, how did you check that `DropWhile` had a request of 1?  As far as compatibility mode, what is the exact process and what am I looking for to know if I've been successful or not?

spadafiva:
@broadwaylamb No problem.  I've reverted to before that merge commit.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,107 @@
...
+                    downstream: subscriber,
+                    decoder: decoder
+                )
+                upstream.receive(subscriber: decodeSubscriber)
+        }
+    }
+}
+
+// swiftlint:disable:next line_length
+private final class _Decode<Upstream: Publisher, Downstream: Subscriber, Coder: TopLevelDecoder>:
```
spadafiva: @broadwaylamb There was no way to get this to fit in the 90 chars.  Does it have to be 90 or can we pick a slightly bigger number?  Otherwise I think ignoring this one rule here is the right thing to do.

spadafiva:
@broadwaylamb Please let me know what else needs to be done to finish this one out.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+//  Created by Joseph Spadafora on 6/21/19.
+//
+
+extension Publishers {
+
+    public struct Decode<Upstream, Output, Coder> : Publisher where
+        Upstream : Publisher,
+        Output : Decodable,
+        Coder : TopLevelDecoder,
+        Upstream.Output == Coder.Input {
```
broadwaylamb: Let's reformat this like in the rest of the codebase:

```swift
public struct Decode<Upstream, Output, Coder>: Publisher
    where Upstream: Publisher,
          Output: Decodable,
          Coder: TopLevelDecoder,
          Upstream.Output == Coder.Input
{
```

I couldn't tune SwiftLint to enforce this kind of formatting :(

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Error
+
+        public let upstream: Upstream
+
+        internal let decoder: Coder
```
broadwaylamb: I can see that this property is private in `Encode`. Probably makes sense to declare it as `private` here too, with leading underscore.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: I believe this is a little bit more readable:

```swift
private final class _Encode<Upstream: Publisher,
                            Downstream: Subscriber,
                            Coder: TopLevelEncoder>
    : OperatorSubscription<Downstream>,
      Subscriber,
      CustomStringConvertible,
      Subscription
    where Coder.Output == Downstream.Input,
          Upstream.Output: Encodable,
          Downstream.Failure == Error
{
```

After some time working on MacBook Pro 13" I came to conclusion that this limit is something you want if you often look at your code using side-by-side view (including git diffs).

13" displays are also most popular among Apple developers.

Swift compiler and standard library follow even shorter line length limit (80 characters), but after some experience using this limit in my own projects I've decided to relax it a bit for convenience, but not too much, to still be able to use side-by-side view.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+        /// This function is called to attach the specified `Subscriber`
+        /// to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<Receiver: Subscriber>(subscriber: Receiver)
+            where Failure == Receiver.Failure, Output == Receiver.Input {
+                let decodeSubscriber = _Decode<Upstream, Receiver, Coder>(
```
broadwaylamb: The body of the function is indented with 8 spaces relative to the function declaration, which is inconsistent with the rest of the project, for example:

https://github.com/broadwaylamb/OpenCombine/blob/8b7bb88d66c4570cb37070d19d8f1f1f8735f8f5/Sources/OpenCombine/Publishers/Publishers.Print.swift#L46-L51

Can you please also rename the generic parameter?

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+    var description: String { return "Decode" }
+
+    init(downstream: Downstream, decoder: Coder) {
+        self._decoder = decoder
+        super.init(downstream: downstream)
+    }
+
+    func receive(subscription: Subscription) {
+        upstreamSubscription = subscription
+        subscription.request(.unlimited)
```
broadwaylamb: Here's how you can test if this request is needed or not:

```swift
// `CustomSubscription` tracks all the requests and cancellations
// in its `history` property
let subscription = CustomSubscription()

// `CustomPublisher` sends the subscription object it has been initialized with
// to whoever subscribed to the `CustomPublisher`.
let publisher = CustomPublisher(subscription: subscription)

// `_Decode` helper will receive the `CustomSubscription `
let decode = publisher.decode(type: [String : String].self, decoder: JSONDecoder())

// This is actually `_Decode`

// `TrackingSubscriber` records every event like "receiveSubscription",
// "receiveValue" and "receiveCompletion" into its `history` property,
// optionally executing the provided callbacks.
let tracking = TrackingSubscriber(
    receiveSubscription: {
        $0.request(.max(42))
        downstreamSubscription = $0
    },
    receiveValue: { _ in .max(2) }
)

drop.subscribe(tracking)

XCTAssertEqual(subscription.history, [.requested(.unlimited)])
```

Let's add this test to the `Encode` and `Decode` test suites, run them in compatibility mode against Apple's Combine and see if it passes. If not, we're gonna need to tweak this test so that it passes in compatibility mode. After that, adjust your implementation accordingly.

I'm going to add some docs for those test helpers.

**UPD**: The docs have arrived 04bbc899566c7f7323ac074294055f44c341f4b2 :)

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+    Downstream.Input: Decodable,
+    Coder.Input == Upstream.Output,
+    Downstream.Failure == Error {
+
+    typealias Input = Upstream.Output
+    typealias Failure = Upstream.Failure
+    typealias Output = Downstream.Input
+
+    private let _decoder: Coder
+    private var _demand: Subscribers.Demand = .none
```
broadwaylamb: This property is not needed, as `Encode`/`Decode` have map-like semantics — they just republish their input by applying some transformation.

In fact, both operators can be implemented in terms of `TryMap`:

```swift
Publishers
    .Just(encoded)
    .decode(type: [Int].self, decoder: JSONDecoder())
// ↑ are equivalent ↓
Publishers
    .Just(encoded)
    .tryMap { try JSONDecoder().decode([Int].self, from: $0) }
```

In the implementation of `Map` we don't change the requested demand in any way, we just re-request it from the upstream.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,108 @@
...
+    func request(_ demand: Subscribers.Demand) {
+        _demand = demand
+    }
+}
+
+extension Publisher {
+    public func decode<Item, Coder>(type: Item.Type, decoder: Coder)
+        -> Publishers.Decode<Self, Item, Coder>
+        where Item : Decodable, Coder : TopLevelDecoder, Self.Output == Coder.Input {
+            return Publishers.Decode(upstream: self, decoder: decoder)
```
broadwaylamb: Let's fix indentation and colons here as well, for example see
https://github.com/broadwaylamb/OpenCombine/blob/8b7bb88d66c4570cb37070d19d8f1f1f8735f8f5/Sources/OpenCombine/Publishers/Publishers.Multicast.swift#L53-L59

Note that the inheritance requirements are declared inside the angle  instead of the `where` clause.

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,109 @@
...
+//
+//  Created by Joseph Spadafora on 6/22/19.
+//
+
+extension Publishers {
+
+    public struct Encode<Upstream, Coder> : Publisher where
+        Upstream : Publisher,
+        Coder : TopLevelEncoder,
+        Upstream.Output : Encodable {
```
broadwaylamb: Can you please address the same comments I've made on `Decode` here too?

Tests/OpenCombineTests/Helpers/TestDecodable.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+import Foundation
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+struct TestDecodable: Codable, Equatable {
```
broadwaylamb: We probably don't need a new type just to test encoding/decoding, we can go with a dictionary `[String : String]` or an array `[Int]`.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,55 @@
...
+        ("testDownstraemReceivesFailure", testDownstreamReceivesFailure)
+    ]
+
+    private let jsonEncoder = JSONEncoder()
+    private let jsonDecoder = JSONDecoder()
+
+    let testValue = TestDecodable()
+
+    func testDecodeWorks() {
+        let data = try! jsonEncoder.encode(testValue)
```
broadwaylamb: I know that this won't crash, but nothing on the type level says so. Crashing the test suite is not good, because it prevents further tests from execution. Let's mark the test method `throws` and remove this exclamation mark.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,55 @@
...
+import XCTest
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+@available(macOS 10.15, *)
+final class DecodeTests: XCTestCase {
```
broadwaylamb, line 17: As I've already mentioned, `Decode` and `Map` have almost the same semantics, which means you can steal [some tests for `Map`/`TryMap`](https://github.com/broadwaylamb/OpenCombine/blob/e3146a1efba678e6a3716ac7dc21958e7def2a78/Tests/OpenCombineTests/PublisherTests/MapTests.swift).

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,55 @@
...
+    let testValue = TestDecodable()
+
+    func testDecodeWorks() {
+        let data = try! jsonEncoder.encode(testValue)
+
+        var decodedValue: TestDecodable?
+        _ = Publishers
+            .Just(data)
+            .decode(type: TestDecodable.self, decoder: jsonDecoder)
+            .sink(receiveValue: { foundValue in
```
broadwaylamb: I think we shouldn't use unrelated entities like `Sink` or `Just` unless there's a real need to do so.

You probably better go with `PassthroughSubject`/`CustomPublisher` and `TrackingSubscriber`, because these are customizable, and allow better test isolation: if there's a regression in `Sink` implementation, we don't want the `Decode` tests to fail as well.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,55 @@
...
+                decodedValue = foundValue
+            })
+
+        XCTAssert(testValue.identifier == decodedValue?.identifier)
+    }
+
+    func testDownstreamReceivesFailure() {
+        var decodeError: Error?
+        let failData = "whoops".data(using: .utf8)!
```
broadwaylamb: You can use the standard library API `Data("whoops".utf8)` and avoid force-unwrapping.

broadwaylamb (changes requested):

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: 👍

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: got it

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: sure

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: I've added this, but this test suite keeps failing for me locally when running .  I think it has to do with JSONEncoder / Decoder. 

```
spadafiva@Spadaputer-Home OpenCombine % swift test -Xswiftc -DOPENCOMBINE_COMPATIBILITY_TEST
[6/6] Linking OpenCombinePackageTests
Test Suite 'All tests' started at 2019-06-30 00:53:11.019
//
// ... all the other tests
//
Test Suite 'DecodeTests' started at 2019-06-30 00:53:11.365
Test Case '-[OpenCombineTests.DecodeTests testDecodeWorks]' started.
Exited with signal code 11
```

Does this happen for you on your machine locally as well?

spadafiva: @broadwaylamb ^

spadafiva (commented):
@broadwaylamb Since Int doesn't encode/decode easily using JSONEncoder/Decoder,  I've also tweaked some of the test helpers to be generic over their value instead of Int.  They are now bounded only by having to conform to Equatable.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: Okay, the semantics make sense to me, but I'm not totally clear on how you'd like this to be changed to work.  Can you explain?

broadwaylamb: Like in the `Map` implementation:

https://github.com/broadwaylamb/OpenCombine/blob/73071a2a9b507239345e781bd041ddd4abe96054/Sources/OpenCombine/Publishers/Publishers.Map.swift#L67-L70

https://github.com/broadwaylamb/OpenCombine/blob/73071a2a9b507239345e781bd041ddd4abe96054/Sources/OpenCombine/Publishers/Publishers.Map.swift#L80-L82

I. e. we just forward the method calls, hence no need to store the demand in this property.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: My output:

```
Test Suite 'DecodeTests' started at 2019-06-30 13:11:12.480
Test Case '-[OpenCombineTests.DecodeTests testDecodeWorks]' started.
/Users/sergej/Developer/tmp/Tests/OpenCombineTests/PublisherTests/DecodeTests.swift:39: error: -[OpenCombineTests.DecodeTests testDecodeWorks] : XCTAssertEqual failed: ("[subscription]") is not equal to ("[subscription, value(["test": "TestDecodable"])]")
Test Case '-[OpenCombineTests.DecodeTests testDecodeWorks]' failed (0.002 seconds).
Test Case '-[OpenCombineTests.DecodeTests testDemand]' started.
/Users/sergej/Developer/tmp/Tests/OpenCombineTests/PublisherTests/DecodeTests.swift:90: error: -[OpenCombineTests.DecodeTests testDemand] : XCTAssertEqual failed: ("[OpenCombineTests.CustomSubscription.Event.requested(Combine.Subscribers.Demand.max(42))]") is not equal to ("[OpenCombineTests.CustomSubscription.Event.requested(Combine.Subscribers.Demand.unlimited)]")
Test Case '-[OpenCombineTests.DecodeTests testDemand]' failed (0.001 seconds).
Test Case '-[OpenCombineTests.DecodeTests testDownstreamReceivesFailure]' started.
Fatal error: Index out of range
Exited with signal code 4
```

What macOS version are you using?

Sources/OpenCombine/Publishers/Publishers.Decode.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    func receive(completion: Subscribers.Completion<Failure>) {
+        switch completion {
+        case .finished:
+            downstream.receive(completion: .finished)
+        case .failure(let error):
+            downstream.receive(completion: .failure(error))
+        }
```
broadwaylamb: I've added a convenience method on `Subscribers.Completion`, you can leverage it here so you don't have to switch:
```swift
return downstream.receive(completion: completion.eraseError())
```

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: This line is indented with 8 spaces instead of 4

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+//  Created by Joseph Spadafora on 6/22/19.
+//
+
+extension Publishers {
+
+    public struct Encode<Upstream, Coder>: Publisher
+        where Upstream : Publisher,
+              Coder : TopLevelEncoder,
+              Upstream.Output : Encodable
```
broadwaylamb: No space before colon in inheritance clause

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,115 @@
...
+        /// This function is called to attach the specified `Subscriber`
+        /// to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<Receiver: Subscriber>(subscriber: Receiver)
+            where Failure == Receiver.Failure, Output == Receiver.Input {
+                let encodeSubscriber = _Encode<Upstream, Receiver, Coder>(
```
broadwaylamb: Let's reindent this too as in `Decode`

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,115 @@
...
+    func request(_ demand: Subscribers.Demand) {
+        _demand = demand
+    }
+}
+
+extension Publisher {
+    public func encode<Coder>(
+        encoder: Coder
+    ) -> Publishers.Encode<Self, Coder>
+        where Coder : TopLevelEncoder
```
broadwaylamb: No space before colon

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        // When
+        publisher.subscribe(subscriber)
+        subject.send(failData)
+
+        // Then
+        guard case .completion(.failure) = subscriber.history[1] else {
+            XCTFail("Decode failure not found")
+            return
+        }
```
broadwaylamb: I have this test crashing in compatibility mode because of index out of bounds error.

Let's compare the whole history for equality:

```swift
// Then
XCTAssertEqual(subscriber.history, [.subscription(Subscriptions.empty), .completion(.failure)])
```

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,94 @@
...
+        let tracking = TrackingSubscriberBase<[String: String], Error>(
+            receiveSubscription: {
+                $0.request(.max(42))
+                downstreamSubscription = $0
+            },
+            receiveValue: { _ in .max(2) }
+        )
+
+        decode.subscribe(tracking)
+        XCTAssert(downstreamSubscription != nil) // Removes unused variable warning
```
broadwaylamb: `XCTAssertNotNil(downstreamSubscription)`

Tests/OpenCombineTests/PublisherTests/EncodeTests.swift
```diff
@@ -0,0 +1,77 @@
...
+        let tracking = TrackingSubscriberBase<Data, Error>(
+            receiveSubscription: {
+                $0.request(.max(42))
+                downstreamSubscription = $0
+            },
+            receiveValue: { _ in .max(2) }
+        )
+
+        encode.subscribe(tracking)
+        XCTAssert(downstreamSubscription != nil) // Removes unused variable warning
```
broadwaylamb: `XCTAssertNotNil(downstreamSubscription)`

Tests/OpenCombineTests/PublisherTests/EncodeTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+        // When
+        publisher.subscribe(subscriber)
+        subject.send(testValue)
+
+        // Then
+        guard case let .value(data) = subscriber.history[1] else {
+            XCTFail("No encoded data found")
+            return
+        }
```
broadwaylamb, line 67: Let's compare the whole history here as well.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: Do other tests work?

```
$ swift test -Xswiftc -DOPENCOMBINE_COMPATIBILITY_TEST --filter OpenCombineTests.DropWhileTests
```

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: Cool, that makes perfect sense

Sources/OpenCombine/Publishers/Publishers.Encode.swift
spadafiva: Sorry about that, I didn't mess with the indentation and they're a little wonky in the Publisher.swift file.

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
spadafiva: For my unit tests in my day to day, I use Quick/Nimble, so I'm less familiar with the regular XCT functions.  Thanks

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
spadafiva: The JSONDecoder is throwing the error here which is why I'm using the `guard case` syntax so that the error comparison doesn't get messy here.

spadafiva: Looks like any decoding error compares to equal here, so this should work

Tests/OpenCombineTests/PublisherTests/EncodeTests.swift
spadafiva: I'm actually trying to get the `data` out of the value here to test for the decoding actually working.  I can add in a separate test for the history though.

Sources/OpenCombine/Publishers/Publishers.Decode.swift
spadafiva: I'm using this version: 10.15 Beta (19A471t).  

Output:
```
g1
g2
g3
Exited with signal code 11
```

code: 

```

    func testDecodeWorks() throws {
        do {
        // Given
        print("g1")
        let data = try jsonEncoder.encode(testValue)
        print("g2")
        let subject = PassthroughSubject<Data, Error>()
        print("g3")
        let publisher = subject.decode(type: [String: String].self, decoder: jsonDecoder)
        print("g4")
        let subscriber = TrackingSubscriberBase<[String: String], Error>()
        print("g5")
        // When
        publisher.subscribe(subscriber)
        print("g6")
        subject.send(data)
        print("g7")
        // Then
        XCTAssertEqual(subscriber.history, [.subscription(Subscriptions.empty),
                                            .value(testValue)])
        print("g8")
        } catch {
            print(error)
        }
    }
```

Sources/OpenCombine/Publishers/Publishers.Encode.swift
broadwaylamb: It's okay, I understand. Those commented out declarations in Publisher.swift are taken from the generated interface from Combine, that's the style they use in generated interfaces.

Sources/OpenCombine/Publishers/Publishers.Encode.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    func receive(completion: Subscribers.Completion<Failure>) {
+        switch completion {
+        case .finished:
+            downstream.receive(completion: .finished)
+        case .failure(let error):
+            downstream.receive(completion: .failure(error))
+        }
```
broadwaylamb: Let's use `eraseError()` here too

Sources/OpenCombine/Publishers/Publishers.Decode.swift
broadwaylamb: This is weird, I can't reproduce it. Can you create a new clean project and run this code in it? Maybe it'll stop in the debugger.

spadafiva:
@broadwaylamb

I've resolved the testing issues for me locally with custom encode/decode objects and fixed the tests. 

Edit: the Travis build is passing, but it's not updating for some reason, maybe the webhook failed?
Edit2: I rebuilt the build and it's now registering here as passing

Tests/OpenCombineTests/PublisherTests/DecodeTests.swift
```diff
@@ -0,0 +1,107 @@
...
+    static let allTests = [
+        ("testDecodeWorks", testDecodeWorks),
+        ("testDownstraemReceivesFailure", testDownstreamReceivesFailure),
+        ("testDemand", testDemand)
+    ]
+
+    var jsonEncoder: TestEncoder = TestEncoder()
+    var jsonDecoder: TestDecoder = TestDecoder()
+
+    override func setUp() {
```
broadwaylamb, line 27: Call `super.setUp()`

Tests/OpenCombineTests/PublisherTests/EncodeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+            receiveSubscription: { $0.request(.unlimited) }
+        )
+
+        // When
+        publisher.subscribe(subscriber)
+        subject.send(testValue)
+
+        // Then
+        guard let testKey = encoder.encoded.first?.key, encoder.encoded.count == 1 else {
+            throw "Could not get testing data from encoding" as TestingError
```
broadwaylamb: Can we use `XCTFail` here instead of throwing? In that case XCTest will show a line number where the error occurred.

broadwaylamb (approved):
LGTM! Thank you for your work.

I’ve left a couple of comments, but they’re minor.

When the conflicts are resolved I’m ready to merge.

PR #9, opened 2019-06-25
Add map operator

M0rtyMerr:


Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  Publishers.Just.swift
```
broadwaylamb: `Publishers.Map.swift`

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -0,0 +1,92 @@
...
+//  Created by Anton Nazarov on 25.06.2019.
+//
+
+extension Publisher {
+
+    /// Transforms all elements from the upstream publisher with a provided closure.
+    ///
+    /// - Parameter transform: A closure that takes one element as its parameter and returns a new element.
+    /// - Returns: A publisher that uses the provided closure to map elements from the upstream publisher to new elements that it then publishes.
+    public func map<T>(_ transform: @escaping (Self.Output) -> T) -> Publishers.Map<Self, T> {
```
broadwaylamb: Let's remove `Self` in `Self.Output` — it's redundant.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,91 @@
...
+final class MapTests: XCTestCase {
+
+    static let allTests = [
+        ("testJust", testJust),
+        ("testEmpty", testEmpty),
+        ("testError", testError),
+        ("testRange", testRange)
+    ]
+
+    func testJust() {
```
broadwaylamb: I suggest that we don't use `Just`, `Empty` and `Fail` publishers here, and use either `PassthroughSubject` or `CustomPublisher` type instead.

The reason is that `Just`, `Empty` and `Fail` have their own specializations for the `map` operator that have nothing to do with the `Map` struct, for instance:
https://github.com/broadwaylamb/OpenCombine/blob/17f9b0a29975106d2f51d5d9b7858c5b1e612019/Sources/OpenCombine/Publishers/Publishers.Just.swift#L203-L205

This means that these tests do not actually test the `Map` type. Those specializations are already tested here:
https://github.com/broadwaylamb/OpenCombine/blob/eab03def585705da0dc33b96763235a3669fa867/Tests/OpenCombineTests/PublisherTests/JustTests.swift#L304-L306

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,91 @@
...
+        map.subscribe(tracking)
+        // Then
+        XCTAssertEqual(tracking.subscriptions.count, 1)
+        XCTAssertEqual(tracking.history, [
+            .subscription(Subscriptions.empty),
+            .completion(Subscribers.Completion<TestingError>.failure(expectedError))
+        ])
+    }
+
+    func testRange() {
```
broadwaylamb, line 57: This is a good test!

If you have time, it'd be great to also add tests that ensure that demand is properly handled — something like this for `DropWhile`:
https://github.com/broadwaylamb/OpenCombine/blob/17f9b0a29975106d2f51d5d9b7858c5b1e612019/Tests/OpenCombineTests/PublisherTests/DropWhileTests.swift#L90-L129

Note that in the DropWhile case the request-demand logic was probably different from Map.

You may want to steal some more tests from the DropWhile suite :)

broadwaylamb (commented):
Thanks @MortyMerr!

Some minor comments regarding tests.

Also please remove the `Map` declaration from the `RemainingCombineInterface.swift` file.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Sure, thanks, didn't noticed it

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Sure. Added few more. Please review

M0rtyMerr:
@broadwaylamb what do you think about the latest updates? :)
I see few people blocked with missing of `.map` so I am ready to apply any changes today

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    ]
+
+    func testEmpty() {
+        // Given
+        let tracking = TrackingSubscriberBase<Never>(receiveSubscription: { $0.request(.unlimited) })
+        let publisher = PassthroughSubject<Int, Never>()
+        // When
+        publisher.map { $0 * 2 }.subscribe(tracking)
+        // Then
+        XCTAssertEqual(tracking.subscriptions.count, 1)
```
broadwaylamb: This assert is redundant, as `tracking.subscriptions` is just a lazy view on `tracking.history`, which is already checked by the assert below.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    func testError() {
+        // Given
+        let expectedError = TestingError.oops
+        let tracking = TrackingSubscriber(receiveSubscription: { $0.request(.unlimited) })
+        let publisher = PassthroughSubject<Int, TestingError>()
+        // When
+        publisher.map { $0 * 2 }.subscribe(tracking)
+        publisher.send(completion: .failure(expectedError))
+        // Then
+        XCTAssertEqual(tracking.subscriptions.count, 1)
```
broadwaylamb: Same here

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        ])
+    }
+
+    func testDemandSubscribe() {
+        // Given
+        let expectedSubscribeDemand = 42
+        let subscription = CustomSubscription()
+        let publisher = CustomPublisher(subscription: subscription)
+        let map = publisher.map { $0 * 2 }
+        let tracking = TrackingSubscriber(receiveSubscription: { $0.request(.max(expectedSubscribeDemand)) })
```
broadwaylamb: Here you test that if a downstream subscriber (`tracking`) requests some elements in the `receiveSubscription` callback, the same number of elements is requested from the upstream publisher.

This is useful, but let's also make sure that if the downstream subscriber doesn't request any demand at all, the upstream publisher won't receive any requests.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,156 @@
...
+            downstreamSubscription = $0
+        })
+        // When
+        map.subscribe(tracking)
+        try XCTUnwrap(downstreamSubscription).cancel()
+        // Then
+        XCTAssertEqual(subscription.history, [.requested(.unlimited), .canceled])
+    }
+
+    func testCancelAlreadyCancelled() throws {
```
broadwaylamb, line 154: This test's name is slightly misleading, I don't see that the subscription is cancelled twice. It makes sense to either update the name of the test (having in mind that `allTests` should also be updated), or add one more test for actual double-cancel.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        let publisher = CustomPublisher(subscription: subscription)
+        let map = publisher.map { $0 * 2 }
+        var downstreamSubscription: Subscription?
+        let tracking = TrackingSubscriber(receiveSubscription: {
+            $0.request(.unlimited)
+            downstreamSubscription = $0
+        })
+        // When
+        map.subscribe(tracking)
+        try XCTUnwrap(downstreamSubscription).cancel()
```
broadwaylamb, line 166: Good use of `XCTUnwrap`!

broadwaylamb (commented):
I like the updates! More minor comments regarding tests incoming.

Note that I've added SwiftLint to the CI system, you may want to rebase onto master and fix the new warnings.

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Deleted

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Deleted

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Added second Cancellation, that was the idea of test

Tests/OpenCombineTests/PublisherTests/MapTests.swift
M0rtyMerr: Implemented additional test case

broadwaylamb (approved):

PR #10, opened 2019-06-25
Publishers.Count

spadafiva:


spadafiva:
@broadwaylamb I'm adding a placeholder for publishers.count, to start working on it

spadafiva:
@broadwaylamb It looks like finishing up count is going to depend on Publishers.Sequence.  You're working on that one currently, right?

broadwaylamb:
@spadafiva how’s Sequence needed?

spadafiva:
Hmm, it actually may not, but I there was no .count extension for publisher defined, so I thought at first glance it required sequence:

https://github.com/broadwaylamb/OpenCombine/blob/eab03def585705da0dc33b96763235a3669fa867/RemainingCombineInterface.swift#L2853

https://github.com/broadwaylamb/OpenCombine/blob/eab03def585705da0dc33b96763235a3669fa867/RemainingCombineInterface.swift#L2853


Looking into the combine interface, I did find this:

```
@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
extension Publisher {

    /// Publishes the number of elements received from the upstream publisher.
    ///
    /// - Returns: A publisher that consumes all elements until the upstream publisher finishes, then emits a single
    /// value with the total number of elements received.
    public func count() -> Publishers.Count<Self>
}
```

Edit:

Okay, I see that there are actually several publishers not implemented in the Publisher.swift file.  That clears up my confusion.  I think I was expecting those to be in RemainingInterfaces.  Thanks!

broadwaylamb:
@spadafiva 

>I think I was expecting those to be in RemainingInterfaces

Yep, my bad. Should probably put 'em back.

broadwaylamb:
Those extenstions for `Publishers.Sequence` declare specializations for various operators that can be implemented more efficiently. I'm going to implement them myself when implementing `Publishers.Sequence`.

For example, we already have such extensions for `Just`, `Once` and `Optional`:
https://github.com/broadwaylamb/OpenCombine/blob/eab03def585705da0dc33b96763235a3669fa867/Sources/OpenCombine/Publishers/Publishers.Optional.swift#L185-L210

spadafiva:
@broadwaylamb That makes sense.  Thanks!

PR #14, opened 2019-06-26
Arrange class hierarchy for Inner helpers

broadwaylamb:
What's in this PR:

- when implementing helper subscription classes (`Inner`) for operators, there was a lot of code duplication. I attempted to extract the common pieces into base classes (@spadafiva you may want to take a look)
- tweaked the `DropWhile` implementation to match Apple's Combine behavior when requesting elements using a `Subscription` instance

### DIFF
--- a/Sources/OpenCombine/Publishers/OperatorSubscription.swift
+++ b/Sources/OpenCombine/Publishers/OperatorSubscription.swift
@@ -1,0 +1,24 @@
+//
+//  OperatorSubscription.swift
+//  
+//
+//  Created by Sergej Jaskiewicz on 26.06.2019.
+//
+
+internal class OperatorSubscription<Downstream: Subscriber>: CustomReflectable {
+    internal var downstream: Downstream
+    internal var upstreamSubscription: Subscription?
+
+    internal var customMirror: Mirror {
+        return Mirror(self, children: EmptyCollection())
+    }
+
+    internal init(downstream: Downstream) {
+        self.downstream = downstream
+    }
+
+    internal func cancel() {
+        upstreamSubscription?.cancel()
+        upstreamSubscription = nil
+    }
+}
--- a/Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
@@ -24,10 +24,8 @@ extension Publishers {
         public func receive<SubscriberType: Subscriber>(subscriber: SubscriberType)
             where Failure == SubscriberType.Failure, Output == SubscriberType.Input
         {
-            let dropWhile = _DropWhile<Upstream, SubscriberType, (Output) -> Bool>(
-                downstream: subscriber, predicate: predicate
-            )
-            upstream.receive(subscriber: dropWhile)
+            let inner = Inner(downstream: subscriber, predicate: catching(predicate))
+            upstream.receive(subscriber: inner)
         }
     }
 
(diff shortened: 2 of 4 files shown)


Sources/OpenCombine/Publishers/OperatorSubscription.swift
```diff
@@ -0,0 +1,24 @@
...
+//  OperatorSubscription.swift
+//  
+//
+//  Created by Sergej Jaskiewicz on 26.06.2019.
+//
+
+internal class OperatorSubscription<Downstream: Subscriber>: CustomReflectable {
+    internal var downstream: Downstream
+    internal var upstreamSubscription: Subscription?
+
```
spadafiva, line 11: I'm looking forward to tweaking my PRs to use this instead of what I've currently done in my other open ones.  I'm wondering if this should conform to `Subscription` also.  It seems like every subclass is saving the upstream subscription and then sending to the downstream.  We could the override that in Operators like `DropWhile` to request 1 instead of a different number.  Do you think there is enough meaningful overlap in the operators to do that?

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
```diff
@@ -46,88 +45,107 @@ extension Publishers {
...
-      CustomReflectable,
       Subscription
-          where Upstream.Output == Downstream.Input
+    where Upstream.Output == Downstream.Input
 {
-
-    typealias Input = Downstream.Input
+    typealias Input = Upstream.Output
     typealias Failure = Upstream.Failure
+    typealias GenericPredicate = (Input) -> Result<Bool, Downstream.Failure>
```
spadafiva: Maybe something like `DropPredicate` would be a more clear name?

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
```diff
@@ -46,88 +45,107 @@ extension Publishers {
...
-        switch completion {
-        case .finished:
-            _downstream.receive(completion: .finished)
-        case .failure(let error):
-            // Safe to force unwrap here, since Downstream.Failure can be
-            // either Upstream.Failure or Error
-            _downstream.receive(completion: .failure(error as! Downstream.Failure))
+    func request(_ demand: Subscribers.Demand) {
+        if genericPredicate == nil {
+            upstreamSubscription?.request(demand)
```
spadafiva, line 127: It's a little unclear to me what's happening here and why.  It could be I don't understand how demand works totally yet.  Is it worth leaving a comment explaining why you don't pass demand on to the upstream when the predicate is `nil`?

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -32,70 +43,121 @@ extension Publishers {
...
+private class _Map<Upstream: Publisher, Downstream: Subscriber>
+    : OperatorSubscription<Downstream>,
+      CustomStringConvertible,
+      Subscription
+{
     typealias Input = Upstream.Output
-    typealias Failure = Downstream.Failure
-    typealias Transform = (Input) -> Downstream.Input
+    typealias Failure = Upstream.Failure
+    typealias GenericTransform = (Input) -> Result<Downstream.Input, Downstream.Failure>
```
spadafiva: maybe `MapTransform` ?

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -32,70 +43,121 @@ extension Publishers {
...
+        where Downstream.Failure == Upstream.Failure
+    {
+        init(downstream: Downstream,
+             transform: @escaping (Upstream.Output) -> Downstream.Input) {
+            super.init(downstream: downstream,
+                       genericTransform: { input in .success(transform(input)) })
+        }
+
+        func receive(completion: Subscribers.Completion<Failure>) {
+            downstream.receive(completion: completion)
```
spadafiva: If `OperatorSubscription` conformed to subscription, we would not have to have this receive here in the Map.Inner.

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -32,70 +43,121 @@ extension Publishers {
...
+    {
+        init(downstream: Downstream,
+             transform: @escaping (Upstream.Output) throws -> Downstream.Input) {
+            super.init(downstream: downstream,
+                       genericTransform: { input in Result { try transform(input) } })
+        }
+
+        func receive(completion: Subscribers.Completion<Failure>) {
+            switch completion {
+            case .finished:
```
spadafiva: It doesn't look like you're transforming the completion here.  Can this just be `downstream.receive(completion: completion)` instead of the switch?

spadafiva (commented):
@broadwaylamb Just had a few comments / questions on this one, but overall I'm looking forward to seeing it merged so that I can pull the changes into my PR branches.  It will definitely make implementing the operators a bit easier.

Sources/OpenCombine/Publishers/OperatorSubscription.swift
broadwaylamb: >Do you think there is enough meaningful overlap in the operators to do that?

I don't think so. This makes sense only for map-like operators, we don't have many of them.

I intentionally haven't provided any implementation of the `Subscription` protocol here to encourage implementors (including myself in the future) to really think about the semantics of the `request(_:)` method for every particular operator.

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
broadwaylamb: IMO adding the "Drop" word won't bring any new information to the reader — we're already dealing with DropWhile.

Sources/OpenCombine/Publishers/Publishers.Map.swift
broadwaylamb: No, the types won't match: the `completion` argument has the type `Subscribers.Completion<Upstream.Failure>`, and the `downstream.receive(completion:)` method expects `Subscribers.Completion<Error>`.

Since user-defined generic types in Swift are invariant relative to their generic parameters,  `Subscribers.Completion<Upstream.Failure>` is not a subtype of `Subscribers.Completion<Error>` and therefore cannot be passed where the latter is expected

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
broadwaylamb: Yeah, it's not obvious at first glance. Will add a comment.

Sources/OpenCombine/Publishers/Publishers.Map.swift
spadafiva: I see, that's a bit annoying that we can see it should be that way but that the type system can't.

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
spadafiva: That's fair enough.  I think I'm just finding the wording `Generic` confusing because I'm expecting it to be more encompassing than a local typealias.

broadwaylamb:
@spadafiva I've addressed your feedback, can you take a look?

spadafiva (approved):
@broadwaylamb Looks good to me.  I think it's a lot clearer to understand and easier to read.

spadafiva:
@broadwaylamb Once you've got it merged in, ping me if you remember so that I can update my PRs based on the changes from this one.

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
```diff
@@ -46,88 +44,132 @@ extension Publishers {
...
-            upstream.receive(subscriber: dropWhile)
+            let inner = Inner(downstream: subscriber, predicate: catching(predicate))
+            upstream.receive(subscriber: inner)
         }
     }
 }
 
-private final class _DropWhile<Upstream: Publisher, Downstream: Subscriber, Predicate>
-    : Subscriber,
+private class _DropWhile<Upstream: Publisher, Downstream: Subscriber>
```
M0rtyMerr, line 55: Hey, great work :)
Could you please make these changes in a separate PR? These changes don't have a relation to `tryMap` operator

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -10,20 +10,31 @@ extension Publisher {
     /// Transforms all elements from the upstream publisher with a provided closure.
     ///
     /// - Parameter transform: A closure that takes one element as its parameter and returns a new element.
-    /// - Returns: A publisher that uses the provided closure to map elements from the upstream publisher to new elements that it then publishes.
```
M0rtyMerr: Hey, great work :)
Why these reformatting changes were made? As I can see code fits swiftlint rules even without these changes. Many unnecessary changes which have no relation to feature (tryMap) complicate code review process. Could u please revert these changes?

Sources/OpenCombine/Publishers/Publishers.Map.swift
```diff
@@ -32,70 +43,120 @@ extension Publishers {
...
+        where Output == SubscriberType.Input, SubscriberType.Failure == Error
+    {
+        let inner = Inner(downstream: subscriber, transform: catching(transform))
+        upstream.receive(subscriber: inner)
+    }
+}
+
+private class _Map<Upstream: Publisher, Downstream: Subscriber>
+    : OperatorSubscription<Downstream>,
+      CustomStringConvertible,
```
M0rtyMerr: What's the reason of moving CustomStringConvertible into class instead of leaving it in the extension?

Sources/OpenCombine/Publishers/OperatorSubscription.swift
MortyMerr: Agreed with @broadwaylamb.

Sources/OpenCombine/Publishers/Publishers.Map.swift
broadwaylamb: I've accidentally configured SwiftLint in such a way that comments were allowed to ignore the line length limit. Fixed that in master.

Sources/OpenCombine/Publishers/Publishers.Map.swift
broadwaylamb: I have a habit of implementing protocol requirements inside a class body instead of an extension if the class is non-final — this leaves the opportunity to override those methods if we need. Yep, in this case we probably won't need to override 'em, but this is a matter of consistency.

Sources/OpenCombine/Publishers/Publishers.DropWhile.swift
broadwaylamb: This PR is now about `Inner` refactoring, for `TryMap` see #15.

#15 depends on this PR, so it's better to merge this first, and review #15 after that.

spadafiva:
@MortyMerr Can you take a look at this so the @broadwaylamb can merge?  I'm holding off on updating some of my PRs until this is resolved so that I don't have to refactor them again later.

M0rtyMerr (approved):

PR #15, opened 2019-06-27
Implement TryMap

broadwaylamb:
What's in this PR:

- tweaked the `Map` implementation a little — the tests were failing in compatibility mode (cc @MortyMerr)
- Implemented `TryMap` with tests. `TryMap` reuses most of the `Map`'s logic, so not many tests were needed.

### DIFF
--- a/RemainingCombineInterface.swift
+++ b/RemainingCombineInterface.swift
@@ -3687,20 +3687,6 @@ extension Publishers.Last : Equatable where Upstream : Equatable {
     public static func == (lhs: Publishers.Last<Upstream>, rhs: Publishers.Last<Upstream>) -> Bool
 }
 
-extension Publishers.Map {
-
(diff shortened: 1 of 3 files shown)


Sources/OpenCombine/Publishers/OperatorSubscription.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  OperatorSubscription.swift
+//  
+//
+//  Created by Sergej Jaskiewicz on 26.06.2019.
+//
+
+internal class OperatorSubscription<Downstream: Subscriber>: CustomReflectable {
```
M0rtyMerr: Just interested - is explicit `internal` required by swiftlint?

broadwaylamb: Yep

https://github.com/broadwaylamb/OpenCombine/blob/8b7bb88d66c4570cb37070d19d8f1f1f8735f8f5/.swiftlint.yml#L37

PR #17, opened 2019-07-01
Convenience init for Just where Output == Void

own2pwn:


broadwaylamb:
Closing because of https://github.com/broadwaylamb/OpenCombine/pull/18#issuecomment-507230872

PR #18, opened 2019-07-01
Convenience sink method for use with Void Output type

own2pwn:


broadwaylamb:
Hello @own2pwn! Thanks for your contribution.

However, we cannot add new public APIs in order to maintain compatibility with Apple's Combine. You can find the up-to-date list of Combine's public APIs in [this gist](https://gist.github.com/broadwaylamb/c2c8550d76b3ff851c4c1dbf0a872e26).

broadwaylamb:
As far as I can see, this code compiles even without the change that you proposed:

```swift
import Combine

_ = PassthroughSubject<Void, Never>().sink {}
```

PR #20, opened 2019-07-03
Adds count implementation and tests

spadafiva:
@broadwaylamb It was a little irritating working from the fork, so I moved it here instead.

### DIFF
--- a/Sources/OpenCombine/Publishers/Publishers.Count.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.Count.swift
@@ -69,9 +69,8 @@ private final class _Count<Upstream: Publisher, Downstream: Subscriber>
     typealias Output = Int
     typealias Failure = Downstream.Failure
 
-    private var _demand: Subscribers.Demand?
-    private var _completion: Subscribers.Completion<Upstream.Failure>?
-
+    private var _hasDemand = false
+    private var _isFinished = false
     private var _count = 0
 
     var description: String { return "Count" }
@@ -83,24 +82,24 @@ private final class _Count<Upstream: Publisher, Downstream: Subscriber>
 
     func receive(_ input: Input) -> Subscribers.Demand {
         _count += 1
-        if _demand == nil {
+        if !_hasDemand {
             return .none
         } else {
-            return _completion != nil ? .unlimited : .none
+            return _isFinished ? .unlimited : .none
         }
     }
 
     func receive(completion: Subscribers.Completion<Upstream.Failure>) {
         if case .finished = completion {
-            _demand = downstream.receive(_count)
-            _completion = .finished// = true
+            _ = downstream.receive(_count)
+            _isFinished = true
         }
         downstream.receive(completion: completion)
     }
 
     func request(_ demand: Subscribers.Demand) {
-        guard _demand == nil else { return }
-        _demand = demand
+        guard !_hasDemand else { return }
+        _hasDemand = true
         if demand > 0 {
             upstreamSubscription?.request(.unlimited)
         } else {
--- a/Sources/OpenCombine/Publishers/Publishers.Count.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.Count.swift
@@ -1,0 +1,99 @@
+//
+//  Publishers.Count.swift
+//
+//
+//  Created by Joseph Spadafora on 6/25/19.
+//
+
+import Foundation
+
+extension Publishers {
+
+    /// A publisher that publishes the number of elements received
+    /// from the upstream publisher.
+    public struct Count<Upstream> : Publisher where Upstream : Publisher {
+
+        /// The kind of values published by this publisher.
+        public typealias Output = Int
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Upstream.Failure
+
+        /// The publisher from which this publisher receives elements.
+        public let upstream: Upstream
+
+        public init(upstream: Upstream) {
+            self.upstream = upstream
+        }
+
+        /// This function is called to attach the specified `Subscriber`
+        /// to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<SubscriberType: Subscriber>(subscriber: SubscriberType)
+            where Upstream.Failure == SubscriberType.Failure,
+                  SubscriberType.Input == Publishers.Count<Upstream>.Output
+        {
+            let count = _Count<Upstream, SubscriberType>(downstream: subscriber)
+            upstream.receive(subscriber: count)
+        }
+    }
+}
+
+extension Publisher {
+
+    /// Publishes the number of elements received from the upstream publisher.
+    ///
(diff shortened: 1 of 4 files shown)


spadafiva:
@broadwaylamb Let me know what changes / fixes you'd like to see on this one.

broadwaylamb:
@spadafiva can you please rebase onto master?

spadafiva:
@broadwaylamb done

spadafiva:
@broadwaylamb and check out this one also?

spadafiva:
@broadwaylamb I've updated the test suite and simplified the implementation here.  This should be ready for a look.

Sources/OpenCombine/Publishers/Publishers.Count.swift
```diff
@@ -0,0 +1,98 @@
...
+    func receive(completion: Subscribers.Completion<Upstream.Failure>) {
+        if case .finished = completion {
+            _ = downstream.receive(_count)
+        }
+        downstream.receive(completion: completion)
+    }
+
+    func request(_ demand: Subscribers.Demand) {
+        guard upstreamSubscription == nil else { return }
+        upstreamSubscription?.request(.unlimited)
```
broadwaylamb: This line will never be executed

Sources/OpenCombine/Publishers/Publishers.Count.swift
```diff
@@ -0,0 +1,98 @@
...
+    public func count() -> Publishers.Count<Self> {
+        return Publishers.Count(upstream: self)
+    }
+}
+
+private final class _Count<Upstream: Publisher, Downstream: Subscriber>
+    : OperatorSubscription<Downstream>,
+    Subscriber,
+    CustomStringConvertible,
+    Subscription
```
broadwaylamb: Let's indent these three lines with 2 more spaces, so that they're aligned with `OperatorSubscription<Downstream>`

Sources/OpenCombine/Publishers/Publishers.Count.swift
```diff
@@ -0,0 +1,98 @@
...
+    typealias Failure = Downstream.Failure
+
+    private var _count = 0
+
+    var description: String { return "Count" }
+
+    func receive(subscription: Subscription) {
+        upstreamSubscription = subscription
+        downstream.receive(subscription: self)
+        upstreamSubscription?.request(.unlimited)
```
broadwaylamb, line 79: If we swap the lines 78 and 79, no test will fail. Can you please write a test that ensures that this calls are in the correct order? The order is important here.

Sources/OpenCombine/Publishers/Publishers.Count.swift
```diff
@@ -0,0 +1,98 @@
...
+        /// This function is called to attach the specified `Subscriber`
+        /// to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<SubscriberType: Subscriber>(subscriber: SubscriberType)
+            where Upstream.Failure == SubscriberType.Failure,
+                  SubscriberType.Input == Publishers.Count<Upstream>.Output
```
broadwaylamb: Let's replace `Publishers.Count<Upstream>.Output` with just `Output` (or even `Int`, if you want).

Tests/OpenCombineTests/PublisherTests/CountTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        }
+        XCTAssertEqual(tracking.history, [.subscription("Count")])
+
+        publisher.send(completion: .finished)
+        XCTAssertEqual(tracking.history, [.subscription("Count"),
+                                          .value(sendAmount),
+                                          .completion(.finished)])
+    }
+
+    func testCountWaitsUntilFinishedToSend() {
```
broadwaylamb, line 53: Isn't this test equivalent to the test above?

Tests/OpenCombineTests/PublisherTests/CountTests.swift
```diff
@@ -0,0 +1,169 @@
...
+                $0.request(.max(42))
+                downstreamSubscription = $0
+            },
+            receiveValue: { _ in .max(4) }
+        )
+
+        countPublisher.subscribe(tracking)
+
+        XCTAssertNotNil(downstreamSubscription)
+        dump(type(of: downstreamSubscription!))
```
broadwaylamb: I've accidentally left this `dump` in one of the test classes — it usually helps to shed some light on Combine internals. Let's remove it.

Tests/OpenCombineTests/PublisherTests/CountTests.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        // Given
+        XCTAssertEqual(subscription.history, [])
+        countPublisher.subscribe(tracking)
+
+        // Then
+        XCTAssertEqual(subscription.history, [.requested(.unlimited)])
+    }
+
+    func testNoDemand() {
```
broadwaylamb: If I'm not mistaken, the two above tests cover the same cases as this test. Maybe remove it?

Sources/OpenCombine/Publishers/Publishers.Count.swift
spadafiva: Oops :)

Sources/OpenCombine/Publishers/Publishers.Count.swift
spadafiva: 👍

Sources/OpenCombine/Publishers/Publishers.Count.swift
spadafiva: Hmmm....That's true.  All the tests still pass if this is an empty function.  That seems odd to me but this passed in compatibility mode.

Tests/OpenCombineTests/PublisherTests/CountTests.swift
spadafiva: The both cover some of the same things, but the one above is focused on checking specifically that the count is correct.  The bottom test is checking that the history doesn't have .value elements added after each send event.

Tests/OpenCombineTests/PublisherTests/CountTests.swift
spadafiva: Got it.

Tests/OpenCombineTests/PublisherTests/CountTests.swift
spadafiva: I'm testing that the "receiveSubscription" and "receiveValue" have different cases in them, so they cover similar ground but are ensuring that there are no differences between different demand requests.  Feel free to add a commit removing either one if you think that's overkill.

Sources/OpenCombine/Publishers/Publishers.Count.swift
spadafiva: Okay, I've added a test to ensure that this is the correct order

spadafiva:
@broadwaylamb What are you using "missing functionality" to track here?  I'm confused by this tag, since the functionality was covered.  Maybe we need a tag for test coverage, or a custom tag for compatibility tests missing or something?

Anyway, this should be ready for another look.

broadwaylamb (approved):

broadwaylamb:
@spadafiva this label is for marking issues and PRs related to the functionality that's present in Combine but missing in OpenCombine. PRs with this label are those that add such functionality.

PR #22, opened 2019-07-04
Implement Filter/TryFilter

spadafiva:


spadafiva:
@broadwaylamb could you take a look at this one as well?

Tests/OpenCombineTests/Helpers/TestHelper.swift
```diff
@@ -0,0 +1,52 @@
...
+import XCTest
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+@available(macOS 10.15, *)
+class TestHelper<Value: Equatable,
```
spadafiva: There is a lot of really similar boilerplate in the tests for setting up the custom subscriptions, custom publisher and tracking subscribers.  This makes setting up a test case as simple as passing a closure to create the chained subscriber.  Let me know what you think @broadwaylamb .  If not exactly this, something along these lines would help remove a lot of the duplicate code in the testing suites.

Tests/OpenCombineTests/Helpers/TrackingSubscriber.swift
```diff
@@ -87,6 +87,12 @@ final class TrackingSubscriberBase<Value: Equatable,
     private let _receiveCompletion: ((Subscribers.Completion<Failure>) -> Void)?
     private let _onDeinit: (() -> Void)?
 
+    var onSubscribe: ((Subscription) -> Void)?
```
spadafiva, line 89: These are some hooks to make it easier to observe changes on a tracking subscriber.  I'm mostly using the `onSubscribe` in the TestHelper init since self can't be captured until after init

broadwaylamb: maybe it's better make existing properties like `_receiveSubscription` internal instead of adding a new property?

Tests/OpenCombineTests/Helpers/TestHelper.swift
broadwaylamb: I think it's a great idea!

Can you add documentation for this helper so it's clear what this class is for and how it can be used?

Let's just indent generic parameters below to be aligned with `Value`, and rename `SUT` to `Sut` to match existing usages of this name.

```swift
class TestHelper<Value: Equatable,
                 Failure,
                 SourceValue: Equatable,
                 SourcePublisher,
                 Sut: Publisher>
    where Sut.Failure == Failure,
          Sut.Output == Value,
          SourcePublisher: CustomPublisherBase<SourceValue>
{
```

Also, `TestHelper` is too generic name, maybe something like OperatorTestHelper will do?

Besides, since we have these same-type constraints `Sut.Failure == Failure` and `Sut.Output == Value`, do we actually need `Value` and `Failure` generic parameters? We can just use type aliases:

```swift
typealias Value = Sut.Output
typealias Failure = Sut.Failure
```

Tests/OpenCombineTests/Helpers/TrackingSubscriber.swift
spadafiva: I've intentionally left the value which returns a demand as a separate property.  In other projects I've worked on, it's helpful to have something which can inspect what's going on without having to return the correct values.  I think having the closure based observers makes it easier to write more targeted tests against specific behaviors without having to worry about changing the preconditions that are set when the underlying TrackingSubscriber is initialized.  What do you think?

Tests/OpenCombineTests/Helpers/TestHelper.swift
spadafiva: 👍

spadafiva:
@broadwaylamb There's 100% test coverage and I've addressed all the feedback.  I'd also like to be able to refactor to use the OperatorTestHelper in more upcoming PRs.  Let me know what/if anything else needs to be done.

Sources/OpenCombine/Publishers/Publishers.Filter.swift
```diff
@@ -0,0 +1,180 @@
...
+    init(downstream: Downstream,
+         description: String,
+         isIncluded: @escaping (Input) throws -> Bool)
+    {
+        self.description = description
+        self._isIncluded = isIncluded
+        super.init(downstream: downstream)
+    }
+
+    let description: String
```
broadwaylamb: Storing `description` is probably not the way to go. Combine uses two different subclasses for `Filter` and `TryFilter`, which share logic in a common base class. It’s possible to make sure of it using `dump(type(of: subscription))` call.

Subclassing also allows us to avoid force casts below.

You can refer to `Map`/`TryMap` implementation to see how we do this.

broadwaylamb (changes requested):

Sources/OpenCombine/Publishers/Publishers.Filter.swift
```diff
@@ -0,0 +1,208 @@
...
+         errorTransform: @escaping (Upstream.Failure) -> Downstream.Failure,
+         isIncluded: @escaping (Input) -> Result<Bool, Downstream.Failure>) {
+        self._isIncluded = isIncluded
+        self.errorTransform = errorTransform
+        super.init(downstream: downstream)
+    }
+
+    func receive(subscription: Subscription) {
+        upstreamSubscription = subscription
+        subscription.request(.unlimited)
```
broadwaylamb: Can you please add a test to check if Apple's Combine also requests `.unlimited` from the upstream upon subscription?

It seems that there is something less straightforward going on: https://forums.swift.org/t/a-uicontrol-event-publisher-example/26215/18

Almost all operators (except `sink` and maybe something else) in Combine are lazy, which means that they don't immediately consume their input, processing it on-demand instead.

spadafiva: @broadwaylamb I’ve added a test and it passing for me locally in compatibility mode.

spadafiva: @broadwaylamb I’ve added a test and it passing for me locally in compatibility mode.

Tests/OpenCombineTests/PublisherTests/FilterTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                                                 .completion(.failure(.oops))])
+    }
+
+    func testFilterSubscriptionDemand() {
+        let helper = OperatorTestHelper(publisherType: CustomPublisher.self) {
+            $0.filter { _ in true }
+        }
+
+        XCTAssertEqual(helper.subscription.history, [.requested(.unlimited)])
+    }
```
broadwaylamb, line 190: The reason it passes like this is because `OperatorTestHelper` always requests `.unlimited`.

If we change this test to

```swift
func testFilterSubscriptionDemand() {
    let helper = OperatorTestHelper(publisherType: CustomPublisher.self,
                                    trackingDemand: .max(3),
                                    receiveValueDemand: .none) {
                                        $0.filter { $0.isMultiple(of: 2) }
                                    }

    XCTAssertEqual(helper.publisher.send(1), .max(1))
    XCTAssertEqual(helper.publisher.send(2), .max(0))
    XCTAssertEqual(helper.publisher.send(3), .max(1))
    XCTAssertEqual(helper.publisher.send(4), .max(0))
    XCTAssertEqual(helper.publisher.send(5), .max(1))
    XCTAssertEqual(helper.publisher.send(6), .max(0))
    XCTAssertEqual(helper.publisher.send(7), .max(1))
    XCTAssertEqual(helper.publisher.send(8), .max(0))

    XCTAssertEqual(helper.subscription.history, [.requested(.max(3))])
}
```

it will pass in compatibility mode, but fail with OpenCombine. The same goes for the test below.

I would suggest to adjust all tests to not request `.unlimited`, since it's not very representative.

Maybe it would be even better for `OperatorTestHelper` initializer arguments to not take default values, instead taking them explicitly to avoid this kind of problems.

It's important to test our publishers with different demands.

broadwaylamb (changes requested):

Tests/OpenCombineTests/PublisherTests/FilterTests.swift
broadwaylamb: You can look at my attempt to handle this in `CompactMap` implementation (#32), which is practically the same as `Filter`.

spadafiva:
@broadwaylamb I've fixed the issues and tests and added tests as per CompactMap.  What else needs to happen for this one?

Sources/OpenCombine/Publishers/Publishers.Filter.swift
```diff
@@ -0,0 +1,213 @@
...
+        switch _isIncluded(input) {
+        case .failure(let error):
+            downstream.receive(completion: .failure(error))
+            cancel()
+            _isFinished = true
+            return .none
+        case .success(let isIncluded):
+            if isIncluded {
+                let newDemand = downstream.receive(input)
+                _unsatisfiedDemand += newDemand - 1
```
broadwaylamb: If `newDemand` is `.max(0)`, this will crash as subtracting from zero demand is not permitted.

Better use two in-place mutations here:

```swift
_unsatisfiedDemand -= 1
_unsatisfiedDemand += newDemand
```

And add a test of course :)

Sources/OpenCombine/Publishers/Publishers.Filter.swift
```diff
@@ -0,0 +1,213 @@
...
+{
+    typealias Input = Upstream.Output
+    typealias Output = Downstream.Input
+    typealias Failure = Upstream.Failure
+
+    private let _isIncluded: (Input) -> Result<Bool, Downstream.Failure>
+    private var _unsatisfiedDemand: Subscribers.Demand = .none
+    private var _isFinished = false
+
+    var errorTransform: (Upstream.Failure) -> Downstream.Failure
```
broadwaylamb: Why is this property needed?

Sources/OpenCombine/Publishers/Publishers.Filter.swift
```diff
@@ -0,0 +1,213 @@
+//
+//  Publishers.Filter.swift
+//  
+//
+//  Created by Joseph Spadafora on 7/3/19.
+//
+
+import Foundation
```
broadwaylamb: We don’t want to depend on Foundation, let’s remove this import

broadwaylamb (changes requested):
I’m sorry for the delay, been quite a busy week.

Sources/OpenCombine/Publishers/Publishers.Filter.swift
spadafiva: This allows use to use this in `receive` so that we don't have to write out a different "receive" for every subclass that is using _Filter and instead maps the error.  Otherwise we have an issues with generic typing that you need to force cast the Failure as! Downstream.Failure.

Sources/OpenCombine/Publishers/Publishers.Filter.swift
spadafiva: Ooops, I think that came with the boilerplate.  Good catch

Tests/OpenCombineTests/PublisherTests/FilterTests.swift
```diff
@@ -0,0 +1,418 @@
...
+                        .value(1),
+                        .completion(.failure(TestingError.oops))])
+    }
+
+    func testFilterSubscriptionDemand() {
+        let helper = OperatorTestHelper(publisherType: CustomPublisher.self,
+                                        initialDemand: .max(3),
+                                        receiveValueDemand: .none) {
+                                            $0.filter { $0.isMultiple(of: 2) }
+                                        }
```
OpenCombineBot: - :no_entry_sign: Closure end should have the same indentation as the line that started it. Expected 8, got 40. (`closure_end_indentation`)

Tests/OpenCombineTests/PublisherTests/OptionalPublisherTests.swift
```diff
@@ -367,4 +368,18 @@ final class OptionalPublisherTests: XCTestCase {
 
         XCTAssertEqual(count, 2)
     }
+
+
```
OpenCombineBot: - :no_entry_sign: Limit vertical whitespace to a single empty line. Currently 2. (`vertical_whitespace`)

PR #24, opened 2019-07-05
Implements replaceNil function

spadafiva:


Tests/OpenCombineTests/PublisherTests/ReplaceNilTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        let nilPublisher = PassthroughSubject<Int?, Never>()
+
+        // When
+        let replaceNil = nilPublisher.replaceNil(with: 42)
+
+        // Then
+        let mirror = Mirror(reflecting: replaceNil)
+        let expected = "Mirror for Map<PassthroughSubject<Optional<Int>, Never>, Int>"
+        XCTAssertEqual(mirror.description, expected)
+    }
```
broadwaylamb: I don't think this test is needed — the return type of `nilPublisher.replaceNil` is fixed on the function signature level.

Tests/OpenCombineTests/PublisherTests/ReplaceNilTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+        // When
+        sut.receive(subscriber: subscriber)
+        nilPublisher.send(output)
+
+        // Then
+        XCTAssertEqual(subscriber.history, [.subscription(Subscriptions.empty),
+                                            .value(output)])
+    }
```
broadwaylamb, line 62: I suggest to combine the above two tests into one test, where multiple values are sent. Something like:

```swift
testReplacesNilElement() {
    // Given
    let passthrough = PassthroughSubject<Int?, Never>()
    let sut = passthrough.replaceNil(with: 42)
    let subscriber = TrackingSubscriberBase<Int, Never>(
        receiveSubscription: {
            $0.request(.unlimited)
        }
    )

    // When
    sut.subscribe(subscriber)
    passthrough.send(1)
    passthrough.send(2)
    passthrough.send(nil)
    passthrough.send(4)
    passthrough.send(5)
    passthrough.send(nil)
    passthrough.send(completion: .finished)

    // Then
    XCTAssertEqual(subscriber.history, [.subscription("PassthroughSubject"),
                                        .value(1),
                                        .value(2),
                                        .value(42),
                                        .value(4),
                                        .value(5),
                                        .value(42),
                                        .completion(.finished)])
}
```

Since this operator trivially uses `Map` under the hood (which already has substantial test coverage), I believe we can be a little lazier when it comes to testing it, and just test the basic behavior.

Sources/OpenCombine/Publishers/Publishers.ReplaceNil.swift
```diff
@@ -0,0 +1,27 @@
...
+// the original Combine interface
+extension Publisher {
+
+    /// Replaces nil elements in the stream with the proviced element.
+    ///
+    /// - Parameter output: The element to use when replacing `nil`.
+    /// - Returns: A publisher that replaces `nil` elements from
+    /// the upstream publisher with the provided element.
+    public func replaceNil<T>(with output: T) -> Publishers.Map<Self, T>
+        where Self.Output == T?
```
broadwaylamb: `Self` is redundant here — let's remove it

Sources/OpenCombine/Publishers/Publishers.ReplaceNil.swift
```diff
@@ -0,0 +1,27 @@
...
+//  
+//
+//  Created by Joseph Spadafora on 7/4/19.
+//
+
+import Foundation
+
+// swiftlint:disable generic_type_name
+// We need to disable this linting rule to maintain the same public api that as
+// the original Combine interface
```
broadwaylamb: I believe generic parameter name is not part of an API at all, since there's no way it can be referenced by name at the call site.

If I'm not mistaken, let's use the same name as we already use here:
https://github.com/broadwaylamb/OpenCombine/blob/065b9819342d2e8ec753609cb3d60a34903b94cd/Sources/OpenCombine/Publishers/Publishers.Sequence.swift#L240-L242

Sources/OpenCombine/Publishers/Publishers.ReplaceNil.swift
```diff
@@ -0,0 +1,27 @@
...
+    ///
+    /// - Parameter output: The element to use when replacing `nil`.
+    /// - Returns: A publisher that replaces `nil` elements from
+    /// the upstream publisher with the provided element.
+    public func replaceNil<T>(with output: T) -> Publishers.Map<Self, T>
+        where Self.Output == T?
+    {
+        return Publishers.Map(upstream: self) { element -> T in
+            element ?? output
+        }
```
broadwaylamb: I think we can use a shorter syntax here:

```swift
return Publishers.Map(upstream: self) { $0 ?? output }
```

just to be consistent with
https://github.com/broadwaylamb/OpenCombine/blob/065b9819342d2e8ec753609cb3d60a34903b94cd/Sources/OpenCombine/Publishers/Publishers.Sequence.swift#L240-L246

broadwaylamb:
I've changed some testing APIs on master, can you please rebase?

Tests/OpenCombineTests/PublisherTests/ReplaceNilTests.swift
spadafiva: That's true, but when we are running in compatibility mode if there is an update that changes how replace nil is implemented to use something different with map, this will catch it.  I don't think there's any harm in keeping it.

Tests/OpenCombineTests/PublisherTests/ReplaceNilTests.swift
spadafiva: I think it makes sense to add another test, but it's a good practice to keep tests small and focused.  The other two tests very simply test the other two paths and would help us find any issues.

Tests/OpenCombineTests/PublisherTests/ReplaceNilTests.swift
broadwaylamb: In this case, you can just explicitly annotate types, so it won't compile if the implementation is ever changed. `Mirror` seems like overkill.

spadafiva: @broadwaylamb I've removed the test

Sources/OpenCombine/Publishers/Publishers.ReplaceNil.swift
broadwaylamb: Can you remove the comment as well?

spadafiva: @broadwaylamb Done.  Should be good to go now

broadwaylamb (approved):

PR #26, opened 2019-07-08
Adds Publishers.Deferred and tests

spadafiva:


Sources/OpenCombine/Publishers/Publishers.Deferred.swift
```diff
@@ -0,0 +1,53 @@
...
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<SubscriberType: Subscriber>(subscriber: SubscriberType)
+            where Failure == SubscriberType.Failure,
+                  Output == SubscriberType.Input
+        {
+            let deferredPublisher = createPublisher()
+            deferredPublisher.receive(subscriber: subscriber)
```
broadwaylamb: We should rather call `deferredPublisher.subscribe(subscriber)` here, since this is what its [documentation](https://developer.apple.com/documentation/combine/publisher/3204756-subscribe) says:

>Always call this function instead of `receive(subscriber:)`.

Tests/OpenCombineTests/PublisherTests/DeferredTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+        XCTAssertEqual(subscription.history, [])
+        XCTAssertFalse(deferredCalled)
+
+        sut.subscribe(tracking)
+
+        XCTAssertEqual(subscription.history, [])
+        XCTAssertTrue(deferredCalled)
+
+        subscription.request(.unlimited)
```
broadwaylamb, line 49: Shouldn't we assert on `subscription.history ` after this call?

Tests/OpenCombineTests/PublisherTests/DeferredTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+        _ = Publishers.Deferred {
+            publisher
+        }
+
+        XCTAssertEqual(subscription.history, [])
+    }
+
+    func testDeferredCreatedAfterSubscription() {
+        var deferredCalled = false
```
broadwaylamb: Le's make sure that the closure is called as many times as `Publishers.Deferred` receives a subscriber, and the publisher is not cached instead. Probably better use a counter instead of a `Bool` variable

Tests/OpenCombineTests/PublisherTests/DeferredTests.swift
```diff
@@ -0,0 +1,60 @@
...
+final class DeferredTests: XCTestCase {
+
+    static let allTests = [
+        ("testDeferredNotCalledBeforeSubscription",
+            testDeferredNotCalledBeforeSubscription),
+        ("testDeferredCreatedAfterSubscription",
+            testDeferredCreatedAfterSubscription)
+    ]
+
+    func testDeferredNotCalledBeforeSubscription() {
```
broadwaylamb: This test is a subset of the test below, maybe remove it?

broadwaylamb (approved):

PR #29, opened 2019-07-09
First where

spadafiva:


spadafiva:
@broadwaylamb I've refactored this one to also use a subclassing approach to be in line with #22

Sources/OpenCombine/Publishers/Publishers.First.swift
```diff
@@ -0,0 +1,288 @@
+//
+//  File.swift
+//  
+//
+//  Created by Joseph Spadafora on 7/8/19.
+//
+
+import Foundation
```
OpenCombineBot: - :no_entry_sign: We don't want to depend on Foundation (`no_foundation_dependency`)

Sources/OpenCombine/Publishers/Publishers.First.swift
```diff
@@ -0,0 +1,288 @@
...
+
+extension Publishers {
+
+    /// A publisher that publishes the first element of a stream, then finishes.
+    public struct First<Upstream: Publisher>:Publisher {
```
OpenCombineBot: - :no_entry_sign: Colons should be next to the identifier of the inheriting type (`inheritance_colon`)

PR #30, opened 2019-07-10
Adds Publishers.Last

spadafiva:


Sources/OpenCombine/Publishers/Publishers.Last.swift
```diff
@@ -0,0 +1,98 @@
+//
+//  Publishers.Last.swift
+//  
+//
+//  Created by Joseph Spadafora on 7/9/19.
+//
+
+import Foundation
```
OpenCombineBot, line 8: - :no_entry_sign: We don't want to depend on Foundation (`no_foundation_dependency`)

broadwaylamb:
I'm going to be taking this over in favor of #76. Huge thanks for your work, @spadafiva!

PR #32, opened 2019-07-10
Implement CompactMap

broadwaylamb:


Sources/OpenCombine/Publishers/Publishers.CompactMap.swift
```diff
@@ -0,0 +1,217 @@
...
+        where Downstream.Failure == Error
+    {
+        func receive(completion: Subscribers.Completion<Upstream.Failure>) {
+            if !_isCompleted {
+                _transform = nil
+                downstream.receive(completion: completion.eraseError())
+            }
+        }
+
+        var description: String { return "TryCompactMap" }
```
spadafiva: @broadwaylamb Can you put the properties at the top above the functions in the bodies of these helper classes?

Sources/OpenCombine/Publishers/Publishers.CompactMap.swift
```diff
@@ -0,0 +1,217 @@
...
+        super.init(downstream: downstream)
+    }
+
+    func receive(subscription: Subscription) {
+        upstreamSubscription = subscription
+        downstream.receive(subscription: self)
+    }
+
+    func receive(_ input: Input) -> Subscribers.Demand {
+        switch _transform?(input) {
```
spadafiva: I think this whole section would be more clear if you used `guard let` to unwrap _transform instead of having the `?` at the end of each case.

spadafiva (changes requested):
This is passing locally for me, but I think there are a few spots in here that I wanted to get your feedback about.

Sources/OpenCombine/Publishers/Publishers.CompactMap.swift
```diff
@@ -0,0 +1,217 @@
...
+    : OperatorSubscription<Downstream>,
+      Subscription
+{
+    typealias Input = Upstream.Output
+    typealias Failure = Upstream.Failure
+    typealias Transform = (Input) -> Result<Downstream.Input?, Downstream.Failure>
+
+    fileprivate var _transform: Transform?
+
+    private var _unsatisfiedDemand: Subscribers.Demand = .none
```
spadafiva: I don't think _unsatisfiedDemand is ever read, only ever written.  It's modified in some functions but not ever returned to anything?

PR #39, opened 2019-07-31
Store newly send value in internal variable inside CurrentValueObject

### LINKED ISSUE #38: CurrentValueSubject is not sending correct value
Hey,

first thanks for this project, I think this can have a huge community impact if we drive it forward and Apple doesn't open-source Combine. I was playing around with it and found a behaviour that I guess is not matching the Apple Combine behaviour.

Take this code:

```swift
let subject = CurrentValueSubject<Int, Never>(0)
subject.send(2)
let c = subject.sink { value in
    print(value)
}
```

Apple's variant is printing `2` but OpenCombine is printing `0`. I tracked down the code and saw that we are not storing the send input in the `CurrentValueSubject`. I changed this behaviour but then a lot of tests were failing. Should I open a PR and we can see there, what the correct approach is?

Edit: I opened a PR and every test except 1 is working, which is the synchronisation test. Would be good if you can take a look.

FranzBusch:
This should fix #38 but I think it breaks the synchronisation test. I tried fixing it and from what I can see everything is locked properly but apparently I am missing something since that test is flaky for me locally. Would be great if you can have a look and tell me what you think about this implementation.


broadwaylamb (approved):
This looks amazing, thank you!

I've restarted the tests, they can be a little flaky indeed.

And can you please address the [SwiftLint warnings](https://travis-ci.org/broadwaylamb/OpenCombine/jobs/565954391)?

MaxDesiatov:
This is slightly off-topic, but I have found that https://www.houndci.com/ (in case you didn't try it yet) works perfectly well for SwiftLint on GitHub projects: it's able to run SwiftLint on its own servers and it posts discovered errors and warnings as PR review comments. I'm not affiliated with that project in any way, just mentioning this because it has saved a lot of time during code review for me. You'd only need to register there (free for public repos) and add `.hound.yml` file with this content to specify your custom SwiftLint rules:

```yml
swiftlint:
  config_file: .swiftlint.yml
```

broadwaylamb:
@MaxDesiatov I'll definitely check this out, thanks!

Tests/OpenCombineTests/CurrentValueSubjectTests.swift
```diff
@@ -310,6 +310,39 @@ final class CurrentValueSubjectTests: XCTestCase {
...
+    func testSubscriptionAfterSend() {
+        // Given
+        let passthrough = Sut(0)
+        let subscriber = TrackingSubscriber(receiveSubscription: { subscription in
+            subscription.request(.unlimited)
+        })
+
+        // When
+        passthrough.send(2)
+        print(passthrough.value)
```
broadwaylamb: Can you please remove this print?

Tests/OpenCombineTests/CurrentValueSubjectTests.swift
```diff
@@ -310,6 +310,39 @@ final class CurrentValueSubjectTests: XCTestCase {
...
+    func testSubscriptionAfterSet() {
+        // Given
+        let passthrough = Sut(0)
+        let subscriber = TrackingSubscriber(receiveSubscription: { subscription in
+            subscription.request(.unlimited)
+        })
+
+        // When
+        passthrough.value = 3
+        print(passthrough.value)
```
broadwaylamb: And this too

Sources/OpenCombine/CurrentValueSubject.swift
```diff
@@ -14,20 +14,33 @@ public final class CurrentValueSubject<Output, Failure: Error>: Subject {
...
+                _lock._unlock()
+            }
+            _lock._lock()
+            return _value
+        }
+        set {
+            defer {
+                _lock._unlock()
+            }
+            _lock._lock()
```
broadwaylamb: I see that the synchronization test failure reproduces every time. Have you tried removing this locking?

FranzBusch:
I removed the print statements and fixed the SwiftLint warning. @MaxDesiatov suggestion for hound is I think a really good one, we used that in our projects before as well. Another solution could be to setup Danger on the CI with that you can also enforce many more conventions that you want to have in your PRs.

Lastly, regarding the test failure. I banged my head a bit against that and then realised it is really easy. The `+=/-=` operation is not thread safe and we cannot synchronise it from within the `CurrentValueSubject`. I quickly verified it against Apples implementation and also there I get synchronisation problems. To fix the test I locked around the += operation. @broadwaylamb What do you think about this?

Sources/OpenCombine/CurrentValueSubject.swift
```diff
@@ -22,16 +24,23 @@ public final class CurrentValueSubject<Output, Failure: Error>: Subject {
 
     /// The value wrapped by this subject, published as a new element whenever it changes.
     public var value: Output {
-        didSet {
-            send(value)
+        get {
+            defer {
+                _lock._unlock()
+            }
+            _lock._lock()
```
broadwaylamb: I'm not sure Combine uses synchronized access to the `value` property though. Maybe that's why the test failed.

Tests/OpenCombineTests/CurrentValueSubjectTests.swift
```diff
@@ -441,13 +478,16 @@ final class CurrentValueSubjectTests: XCTestCase {
...
         race(
             {
+                lock.lock()
                 cvs.value += 1
+                lock.unlock()
             },
             {
+                lock.lock()
                 cvs.value -= 1
+                lock.unlock()
```
broadwaylamb: Maybe we should replace `cvs.value += 1` with just `cvs.send(42)`, and also remove this locking? It feels like it makes this test useless, since the access is guaranteed to be synchronized at test site, and we want to test the synchronization of `CurrentValueSubject` itself.

broadwaylamb:
@FranzBusch also can you please fix the conflicts so it can be merged?

Sources/OpenCombine/CurrentValueSubject.swift
FranzBusch: I would guess that they at synchronise the access to value otherwise this would not be thread safe at all. The test failed because what tested is not something the the Combine/OpenCombine framework can synchronise but rather the user of the framework needs to synchronise.

Tests/OpenCombineTests/CurrentValueSubjectTests.swift
FranzBusch: I removed the lock and just send 42 now. The thread sanitiser should catch if we have a sync problem here I guess.

FranzBusch:
@broadwaylamb Conflicts are resolved.

Sources/OpenCombine/CurrentValueSubject.swift
broadwaylamb: I've just checked the assembly, access to `value` is indeed not thread-safe.

Tests/OpenCombineTests/CurrentValueSubjectTests.swift
broadwaylamb: Thanks! Does the sanitizer complain if you remove the lock from the `value` getter and setter?

Sources/OpenCombine/CurrentValueSubject.swift
FranzBusch: Hm, I tested the value in access in Combine and at least the thread sanitiser doesn't complain. What do you want me to do here?

FranzBusch: I removed the locks from within the get and set and the sanitiser didn't complain. So I removed it now.

broadwaylamb: In this case, I think I'm ready to merge it, just waiting for you to push

FranzBusch:
@broadwaylamb Done go ahead :)

broadwaylamb:
@FranzBusch will you remove that lock?

FranzBusch:
@broadwaylamb Ah sorry, though I pushed. Now :)

Sources/OpenCombine/CurrentValueSubject.swift
```diff
@@ -22,16 +24,19 @@ public final class CurrentValueSubject<Output, Failure: Error>: Subject {
 
     /// The value wrapped by this subject, published as a new element whenever it changes.
     public var value: Output {
-        didSet {
-            send(value)
+        get {
+            return _value
+        }
+        set {
+            send(newValue)
```
broadwaylamb: The backing is not set here. Maybe, since we don’t synchronize anymore, we can revert this to just a stored property with a `didSet` like it was before?

PR #41, opened 2019-08-01
Use Danger CI

broadwaylamb:


Tests/OpenCombineTests/Helpers/CustomSubscription.swift
```diff
@@ -71,3 +71,4 @@ final class CustomSubscription: Subscription, CustomStringConvertible {
 
     var description: String { return "CustomSubscription" }
 }
+
```
OpenCombineBot: - :no_entry_sign: Files should have a single trailing newline. (`trailing_newline`)

PR #44, opened 2019-08-16
Implement IgnoreOutput

epatey:
Thanks for the important project. It will make transitioning to Combine realistic.

To get my feet wet, I took a stab at IgnoreOutput since it's so trivial. Happy to pick up some more once this one goes in.

### DIFF
--- a/Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
+++ b/Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
@@ -1,0 +1,160 @@
+//
+//  IgnoreOutputTests.swift
+//
+//  Created by Eric Patey on 16.08.20019.
+//
+
+import XCTest
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+@available(macOS 10.15, iOS 13.0, *)
+final class IgnoreOutputTests: XCTestCase {
+    static let allTests = [
+        ("testCompletionWithEmpty", testCompletionWithEmpty),
+        ("testCompletionWithValues", testCompletionWithValues),
+        ("testCompletionWithError", testCompletionWithError),
+        ("testDemand", testDemand),
+        ("testSendsSubcriptionDownstreamBeforeDemandUpstream",
+         testSendsSubcriptionDownstreamBeforeDemandUpstream),
+        ("testTestSuiteIncludesAllTests", testTestSuiteIncludesAllTests),
+    ]
+
+    func testCompletionWithEmpty() {
+        let subscription = CustomSubscription()
+        let publisher = CustomPublisher(subscription: subscription)
+        let ignoreOutputPublisher = publisher.ignoreOutput()
+        let tracking = TrackingSubscriberBase<Never, TestingError>(
+            receiveSubscription: { $0.request(.max(42)) }
+        )
+
+        XCTAssertEqual(tracking.history, [])
+
+        ignoreOutputPublisher.subscribe(tracking)
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput")])
+
+        publisher.send(completion: .finished)
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput"),
+                                          .completion(.finished)])
+    }
+
+    func testCompletionWithValues() {
+        let upstreamSubscription = CustomSubscription()
+        let upstreamPublisher = CustomPublisher(subscription: upstreamSubscription)
+        let ignoreOutputPublisher = upstreamPublisher.ignoreOutput()
+        let tracking = TrackingSubscriberBase<Never, TestingError>(
+            receiveSubscription: { $0.request(.max(42)) }
+        )
+
+        XCTAssertEqual(tracking.history, [])
+
+        ignoreOutputPublisher.subscribe(tracking)
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput")])
+
+        XCTAssertEqual(upstreamPublisher.send(666), .none)
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput")])
+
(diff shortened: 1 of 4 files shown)


broadwaylamb:
Thank you @epatey! At first sight this looks good, I'll take a closer look in a couple of days.

Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        let tracking = TrackingSubscriberBase<Never, TestingError>(
+            receiveSubscription: { $0.request(.max(42)) }
+        )
+
+        XCTAssertEqual(tracking.history, [])
+
+        ignoreOutputPublisher.subscribe(tracking)
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput")])
+
+        for _ in 0..<10 { _ = publisher.send(666) }
```
broadwaylamb: Let's add `XCTAssertEqual` here instead of ignoring the return value to ensure that the underlying subscriber correctly handles demand:

```swift
for _ in 0 ..< 10 {
    XCTAssertEqual(publisher.send(666), .none)
}
```

Same applies in tests below.

I would also remove the loop, it seems unnecessary, but that's noncritical and up to you.

Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        XCTAssertEqual(tracking.history, [.subscription("IgnoreOutput"),
+                                          .completion(.failure(.oops))])
+    }
+
+    func testDemand() {
+        // demand from downstream is ignored since no values are ever
+        // sent. upstream demand is set to unlimited and left alone.
+        let subscription = CustomSubscription()
+        let publisher = CustomPublisher(subscription: subscription)
+        let countPublisher = publisher.ignoreOutput()
```
broadwaylamb: Nit: this should be `ignoreOutputPublisher`

Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
```diff
@@ -0,0 +1,159 @@
...
+    func testSendsSubcriptionDownstreamBeforeDemandUpstream() {
+        let sentDemandRequestUpstream = "Sent demand request upstream"
+        let sentSubscriptionDownstream = "Sent subcription downstream"
+        var receiveOrder: [String] = []
+
+        let subscription = CustomSubscription(onRequest: { _ in
+            receiveOrder.append(sentDemandRequestUpstream)
+        })
+        let publisher = CustomPublisher(subscription: subscription)
+        let countPublisher = publisher.count()
```
broadwaylamb: You're testing `Publishers.Count` here :)

broadwaylamb (changes requested):

Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
epatey: I'll get rid of the loop.

To be honest, I am pretty confused about what the `Demand` result from `.receive(input:)` means semantically. The doc says:
> A `Demand` instance indicating how many more elements the subcriber expects to receive.

Why, then, would `IgnoreOutput` tell the upstream that it expects no more values? Maybe the wording that Apple chose isn't precise enough. When you look at the doc for the initial `Demand` given to the `Subscription`, it says:
> Tells a publisher that it may send more values to the subscriber.

So, in my mind, a `Subscription` must keep track of the current demand of the downstream `Subscriber` - initializing it with the value passed into `.request(demand:)` and updating it with the value returned by `.receive(input:)`.

So, given that logic, I would expect `IgnoreOutput` to return `.unlimited` rather than `.none` from `.receive(input:)`.

broadwaylamb: We return `.none` because we requested `.unlimited` in the `receive(subscription:)` method, there's no point in requesting more as we've already requested the maximum possible value.

The `Demand` value returned from `receive(_:)` method may be useful if you request some finite value and then you realize you want more. See, for example, `Publishers.Filter` implementation. Filter republishes the values from upstream unless they don't satisfy a predicate. If the filter's downstream requested e. g. 42 values, it's the filter's responsibility to attempt to provide that number of values to its subscriber. However, if a value received by filter from its upstream does not satisfy the predicate, filter requests exactly one more value from the upstream by returning `.max(1)`, hoping that the next value will satisfy the predicate and filter will be able to satisfy its subscriber's demand.

So yeah, probably the docs put it in a confusing way.

epatey: Ah.  🧠💡I see. It's additive and not absolute. That's make so much more sense to me. D'oh.

Tests/OpenCombineTests/PublisherTests/IgnoreOutputTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        let sentSubscriptionDownstream = "Sent subcription downstream"
+        var receiveOrder: [String] = []
+
+        let subscription = CustomSubscription(onRequest: { _ in
+            receiveOrder.append(sentDemandRequestUpstream)
+        })
+        let publisher = CustomPublisher(subscription: subscription)
+        let ignoreOutputPublisher = publisher.ignoreOutput()
+        let tracking = TrackingSubscriberBase<Never, TestingError>(receiveSubscription: {
+            _ in receiveOrder.append(sentSubscriptionDownstream)
```
OpenCombineBot: - :no_entry_sign: Closure parameters should be on the same line as opening brace. (`closure_parameter_position`)

broadwaylamb (approved):

epatey:
@broadwaylamb, I think this one's ready to go. Let me know if there are any other changes you're looking for.

broadwaylamb:
Thank you!

PR #45, opened 2019-08-17
FlatMap

epatey:
I'm taking a stab at `FlatMap`. In support of operators that need serialization of sending values from multiple upstream/child subscriptions, I've implemented a simple/fast `SerializedWorkQueue` that adheres to the `Combine` threading rules and avoids unnecessary context switches.

Remaining work:
- [x] Clean up and ensure all completion/failures paths are correct.
  - [x] Clean up and complete with `.failed` if either upstream or any children fail.
  - [x] Complete with `.finished` only after upstream and all children have completed (i.e. add refcounting of active subscriptions) AND all buffer values have been sent.
- [x] Figure out and add support for `maxPublishers`
- [x] Figure out non-trivial handling of demand for child subscriptions and implement it.
  - [x] Basic child demand scenarios work
  - [x] Add support for downstream demand going from `.unlimited` to `.max(x)`
  - [x] Add support for downstream demand going from `.max(x)` to `.unlimited`
- [x] Add unit tests 
- [x] Factor out common base class from `Inner` class to be shared with all _merge_ like operators.

### DIFF
--- a/Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
@@ -1,0 +1,409 @@
+//
+//  Publishers.FlatMap.swift
+//
+//  Created by Eric Patey on 16.08.2019.
+//
+
+extension Publisher {
+
+    /// Transforms all elements from an upstream publisher into a new or existing
+    /// publisher.
+    ///
+    /// `flatMap` merges the output from all returned publishers into a single stream of
+    /// output.
+    ///
+    /// - Parameters:
+    ///   - maxPublishers: The maximum number of publishers produced by this method.
+    ///   - transform: A closure that takes an element as a parameter and returns a
+    ///   publisher that produces elements of that type.
+    /// - Returns: A publisher that transforms elements from an upstream publisher into
+    /// a publisher of that element’s type.
+    public func flatMap<Result, Child>(maxPublishers: Subscribers.Demand = .unlimited,
+                                       _ transform: @escaping (Self.Output) -> Child)
+        -> Publishers.FlatMap<Child, Self>
+        where Result == Child.Output, Child: Publisher, Self.Failure == Child.Failure {
+            return Publishers.FlatMap(upstream: self,
+                                      maxPublishers: maxPublishers,
+                                      transform: transform)
+    }
+}
+
+extension Publishers {
+    public struct FlatMap<Child, Upstream>: Publisher
+        where Child: Publisher, Upstream: Publisher, Child.Failure == Upstream.Failure {
+
+        /// The kind of values published by this publisher.
+        public typealias Output = Child.Output
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Upstream.Failure
+
+        public let upstream: Upstream
+
+        public let maxPublishers: Subscribers.Demand
+
+        public let transform: (Upstream.Output) -> Child
+
+        public init(upstream: Upstream, maxPublishers: Subscribers.Demand,
+                    transform: @escaping (Upstream.Output) -> Child) {
+            self.upstream = upstream
+            self.maxPublishers = maxPublishers
+            self.transform = transform
+        }
+
+        /// This function is called to attach the specified `Subscriber` to this
+        /// `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<Downstream>(subscriber: Downstream)
+            where Downstream: Subscriber,
+            Child.Output == Downstream.Input,
+            Upstream.Failure == Downstream.Failure {
+                let inner = Inner(downstream: subscriber,
+                                  maxPublishers: maxPublishers,
+                                  transform: transform)
+                upstream.subscribe(inner)
+        }
+    }
+}
+
+extension Publishers.FlatMap {
+    fileprivate final class Inner<Downstream: Subscriber>
+        : CustomStringConvertible,
+        Cancellable,
+        CustomReflectable
+    where Downstream.Input == Child.Output, Downstream.Failure == Upstream.Failure {
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        private typealias PendingValue = (
+            value: Downstream.Input,
+            // If the value was buffered at the time it became available, and the child's
+            // demand was left at `.none` we keep track of the child in `pausedChild` so
+            // that we can demand some more of it after sending this value.
+            pausedChild: ChildSubscriber?)
+
+        private let lock = Lock(recursive: false)
+        private let maxPublishers: Subscribers.Demand
+        private let transform: (Upstream.Output) -> Child
+        // Locking rules for this class.
+        //  - All mutable state must only be accessed while `lock` is held.
+        //  - In order to avoid any deadlock potential, it is absolutely forbidden to have
+        //      any sort of call out from this class while the lock is held. This is why
+        //      the draining of the work queue uses a relatively complex pattern.
+        private var downstream: Downstream?
+        private var childSubscribers = Set<ChildSubscriber>()
+        private var downstreamDemand = Subscribers.Demand.unlimited
+        private var valuesToSend = [PendingValue]()
+        private var queueIsBeingProcessed = false
+        private var sendFinishedAfterDrainingQueue = false
+        private var upstreamSubscription: Subscription?
+
+        var description: String { return "FlatMap" }
+
+        internal var customMirror: Mirror {
+            return Mirror(self, children: EmptyCollection())
+        }
+
+        init(downstream: Downstream,
+             maxPublishers: Subscribers.Demand,
+             transform: @escaping (Upstream.Output) -> Child) {
+            self.downstream = downstream
+            self.maxPublishers = maxPublishers
+            self.transform = transform
+        }
+
+        final func cancel() {
+            let (upstreamToCancel, childrenToCancel) = lock.do { ()
+                -> (Subscription?, Set<ChildSubscriber>) in
+                let upstreamToCancel = upstreamSubscription
+                upstreamSubscription = nil
+                return (upstreamToCancel, lockedDeactivateAndReturnChildToCancel())
+            }
+            upstreamToCancel?.cancel()
+            cancelChildren(childrenToCancel)
+        }
+    }
+}
+
+// Private implementation
+extension Publishers.FlatMap.Inner {
+    private func deactivate() {
+        cancelChildren(lock.do(lockedDeactivateAndReturnChildToCancel))
+    }
+
+    // Must be called with lock held.
+    private func lockedDeactivateAndReturnChildToCancel() -> Set<ChildSubscriber> {
+        downstream = nil
+        downstreamDemand = .none
+        let result = childSubscribers
+        childSubscribers.removeAll()
+        upstreamSubscription = nil
+        return result
+    }
+
+    private func cancelChildren(_ childrenToCancel: Set<ChildSubscriber>) {
+        childrenToCancel.forEach { $0.cancel() }
+    }
+
+    /// In a thread-safe way, this function performs the passed in work with the lock held
+    /// and then checks to see if either upstream or any of the child subscriptions remain
+    /// active. If there are no remaining active subscriptions, it enqueues the sending
+    /// of `.finished` downstream using the processing queue.
+    /// - Parameter lockedWork: block to be formed with the lock held.
+    private final func maybeSendFinishedAfterExecutingWork(lockedWork: () -> Void) {
+        let shouldProcessQueue: Bool = lock.do {
+            lockedWork()
+            if childSubscribers.isEmpty && upstreamSubscription == nil {
+                sendFinishedAfterDrainingQueue = true
+                if !queueIsBeingProcessed {
+                    queueIsBeingProcessed = true
+                    return true
+                }
+            }
+            return false
+        }
+
+        if shouldProcessQueue {
+            processQueue()
+        }
+    }
+
+    private func receivedCompletion(_ completion: Subscribers.Completion<Failure>,
+                                    fromChild child: ChildSubscriber) {
+        switch completion {
+        case .finished:
+            removeActiveSubscription(forChild: child)
+        case .failure:
+            downstream?.receive(completion: completion)
+            deactivate()
+        }
+    }
+
+    private func removeActiveSubscription(forChild child: ChildSubscriber) {
+        maybeSendFinishedAfterExecutingWork { childSubscribers.remove(child) }
+    }
+
+    private func receivedValue(_ value: Child.Output,
+                               fromChild child: ChildSubscriber) -> Subscribers.Demand {
+        // When receiving a value from a child, we need to determine what additional
+        // demand to return to the child. Apple's logic for this determination is as
+        // follows:
+        //  - If we are in `.unlimited` mode, we always request `.none` additional
+        //      else
+        //  - If there is a surplus relative to the demand, we request `.none`
+        //      else
+        //  - There is not yet a surplus, so request `.max(1)` more from the child
+
+        let (surplusAvailable, processTheQueue): (Bool, Bool) = lock.do {
+            // If we already have enough values to satisfy the demand, we "buffer" this
+            // child value establishing a surplus.
+            if downstreamDemand <= valuesToSend.count {
+                valuesToSend.append((value, child))
+                return (surplusAvailable: true, processTheQueue: false)
+            } else {
+                valuesToSend.append((value, nil))
+                if queueIsBeingProcessed {
+                    return (surplusAvailable: false, processTheQueue: false)
+                }
+                queueIsBeingProcessed = true
+                return (surplusAvailable: false, processTheQueue: true)
+            }
+        }
+
+        let demandResult = surplusAvailable || demandForChild() == .unlimited
+            ? Subscribers.Demand.none : Subscribers.Demand.max(1)
+
+        if processTheQueue {
+            processQueue()
+        }
+
+        return demandResult
+    }
+
+    private func demandForChild() -> Subscribers.Demand {
+        return self.downstreamDemand == .unlimited ? .unlimited : .max(1)
+    }
+
+    private enum QueueWorkStatus {
+        case noWork
+        case sendFinish
+        case sendValues(values: ArraySlice<PendingValue>)
+    }
+
+    private func processQueue() {
+        assert(queueIsBeingProcessed)
+
+        // We loop processing the queue in case somebody put stuff on the queue while we
+        // were sending values with the lock unlocked.
+        while true {
+            let work: QueueWorkStatus = lock.do {
+                if downstreamDemand == .none || valuesToSend.isEmpty {
+                    if sendFinishedAfterDrainingQueue && valuesToSend.isEmpty {
+                        return .sendFinish
+                    } else {
+                        queueIsBeingProcessed = false
+                        return .noWork
+                    }
+                }
+
+                let countToSend = min(valuesToSend.count, downstreamDemand.max ?? .max)
+                let result = valuesToSend[0..<countToSend]
+                // TODO: Consider an alternative storage to avoid O(n) removeFirst
+                valuesToSend.removeFirst(countToSend)
+                downstreamDemand -= countToSend
+                return .sendValues(values: result)
+            }
+
+            guard let downstream = downstream else { return }
+
+            switch work {
+            case .noWork:
+                return
+            case .sendFinish:
+                downstream.receive(completion: .finished)
+                deactivate()
+                return
+            case .sendValues(let values):
+                var newDemand = Subscribers.Demand.none
+                values.forEach {
+                    newDemand += downstream.receive($0.value)
+                    // pausedChild is present only if the value was buffered and the
+                    // child's demand was left at `.none`. In that case, once we send the
+                    // buffered value, we need to tell the child to get another value.
+                    $0.pausedChild?.request(.max(1))
+                }
+
+                if newDemand != .none {
+                    lock.do { downstreamDemand += newDemand }
+                }
+            }
+        }
+    }
+}
+
+// This `Subscriber` implementation is for `FlatMap`'s upstream subscription
+extension Publishers.FlatMap.Inner: Subscriber {
+    // FlatMap received the upstream subscription
+    fileprivate func receive(subscription: Subscription) {
+        upstreamSubscription = subscription
+        downstream?.receive(subscription: self)
+        subscription.request(maxPublishers)
+    }
+
+    /// Receive a new value from the upstream subscription. A new child subscription
+    /// will be made on the `Child` that the input value is transformed into.
+    /// - Parameter input: a value to be transformed by `transform`
+    fileprivate func receive(_ input: Input) -> Subscribers.Demand {
+        let newChildSubscriber = ChildSubscriber(parent: self)
+
+        lock.do { _ = childSubscribers.insert(newChildSubscriber) }
+
+        self.transform(input).subscribe(newChildSubscriber)
+
+        return .none
+    }
+
+    // Upstream subscription completed
+    fileprivate func receive(completion: Subscribers.Completion<Failure>) {
+        switch completion {
+        case .finished:
+            maybeSendFinishedAfterExecutingWork { upstreamSubscription = nil }
+        case .failure:
+            downstream?.receive(completion: completion)
+            deactivate()
+        }
+    }
+}
+
+// Inner is the `Subscription` for `Downstream`
+extension Publishers.FlatMap.Inner: Subscription {
+    fileprivate func request(_ demand: Subscribers.Demand) {
+        let (drainTheQueue, becameUnlimited) = lock.do { () -> (Bool, Bool) in
+            let becameUnlimited =
+                (demand == .unlimited) && (downstreamDemand != .unlimited)
+            downstreamDemand = demand
+            defer { queueIsBeingProcessed = true }
+            return (!queueIsBeingProcessed, becameUnlimited)
+        }
+
+        if becameUnlimited {
+            // TODO: This code isn't yet thread safe. The correct change is to do this
+            // through the queue just like sending values and finished. Finished is
+            // done through the queue as a bit of a hack. The right design is to have
+            // an enum of actions on the queue. That enum will include (send value,
+            // send finished, set child demand).
+            let newChildDemand = demandForChild()
+            childSubscribers.forEach { $0.request(newChildDemand) }
+        }
+
+        if drainTheQueue {
+            processQueue()
+        }
+    }
+}
+
+extension Publishers.FlatMap.Inner {
+    /// ChildSubscriber is needed to help implement the backpressure/demand strategy.
+    /// Specifically, a custom subscriber is needed to manage the demand of the child
+    /// subscription:
+    ///  - Send .max(1) request when the subscription is received
+    ///  - Send .max(1) request when downstream subscriber demands more and a previously
+    ///   buffered value from the child was sent. (When the value was buffered, the
+    ///   child's demand reached .none - effectively pausing the child.)
+    fileprivate final class ChildSubscriber: Hashable {
+        internal typealias Input = Downstream.Input
+        internal typealias Failure = Downstream.Failure
+
+        private var _upstreamSubscription: Subscription?
+        private unowned let _parent: Publishers.FlatMap<Child, Upstream>.Inner<Downstream>
+
+        init(parent: Publishers.FlatMap<Child, Upstream>.Inner<Downstream>) {
+            _parent = parent
+        }
+
+        fileprivate func request(_ demand: Subscribers.Demand) {
+            _upstreamSubscription?.request(demand)
+        }
+
+        public static func == (lhs: ChildSubscriber, rhs: ChildSubscriber) -> Bool {
+            return ObjectIdentifier(lhs) == ObjectIdentifier(rhs)
+        }
+
+        public func hash(into hasher: inout Hasher) {
+            hasher.combine(ObjectIdentifier(self))
+        }
+    }
+}
+
+extension Publishers.FlatMap.Inner.ChildSubscriber: Cancellable {
+    internal func cancel() {
+        _upstreamSubscription?.cancel()
+        _upstreamSubscription = nil
+    }
+}
+
+extension Publishers.FlatMap.Inner.ChildSubscriber: Subscriber {
+    internal func receive(subscription: Subscription) {
+        if _upstreamSubscription == nil {
+            _upstreamSubscription = subscription
+            subscription.request(_parent.demandForChild())
+        } else {
+            assertionFailure()
+            subscription.cancel()
+        }
+    }
+
+    fileprivate func receive(_ input: Input) -> Subscribers.Demand {
+        return _parent.receivedValue(input, fromChild: self)
+    }
+
+    fileprivate func receive(completion: Subscribers.Completion<Failure>) {
+        _parent.receivedCompletion(completion, fromChild: self)
+    }
+}
--- a/Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
@@ -75,8 +75,7 @@ extension Publishers {
 extension Publishers.FlatMap {
     fileprivate final class Inner<Downstream: Subscriber>
         : CustomStringConvertible,
-        Cancellable,
-        CustomReflectable
+        Cancellable
     where Downstream.Input == Child.Output, Downstream.Failure == Upstream.Failure {
         typealias Input = Upstream.Output
         typealias Failure = Upstream.Failure
@@ -105,10 +104,6 @@ extension Publishers.FlatMap {
         private var upstreamSubscription: Subscription?
 
         var description: String { return "FlatMap" }
-
-        internal var customMirror: Mirror {
-            return Mirror(self, children: EmptyCollection())
-        }
 
         init(downstream: Downstream,
              maxPublishers: Subscribers.Demand,
--- a/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
+++ b/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
@@ -346,7 +346,8 @@ final class FlatMapTests: XCTestCase {
                     received777Sem.signal()
                 }
                 return .none
-        })
+            }
+        )
 
         flatMap.subscribe(downstreamSubscriber)
 
--- a/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
+++ b/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
@@ -1,0 +1,624 @@
+//
+//  FlatMapTests.swift
+//
+//  Created by Eric Patey on 17.08.2019.
+//
+
+import XCTest
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+/// Helper function for predictably testing concurrency/race scenarios.
+/// - Parameter block: block to execute concurrently
+///
+/// Apple, surprisingly, calls out to subscribers with a lock held. This will absolutely
+/// block children who send values concurrently until the current downstream value
+/// delivery has been completed.
+///
+/// This means that, without a timeout, the code below is guaranteed to deadlock. Because
+/// of this we need to choose a timeout that is low enough to not materially slow down the
+/// tests, but long enough to ensure that we are effectively testing the desired race
+/// conditions. It needs to be a long enough timeout to allow the caller's block to begin
+/// executing.
+///
+/// I understand that timeouts like this are a smell. I'd be happy to entertain other ways
+/// to deterministically test concurrency/race conditions.
+
+func performConcurrentBlock(_ block: @escaping () -> Void) {
+    let sem = DispatchSemaphore(value: 0)
+    DispatchQueue.global(qos: .background).async {
+        block()
+        sem.signal()
+    }
+    #if OPENCOMBINE_COMPATIBILITY_TEST
+    // If running in compatibility mode, assert that we got a timeout. If not, Apple
+    // changed their implementation to not call out with a lock held.
+    XCTAssertEqual(sem.wait(timeout: DispatchTime.now() + 0.01), .timedOut)
+    #else
+    sem.wait()
+    #endif
+}
+
+@available(macOS 10.15, iOS 13.0, *)
+final class FlatMapTests: XCTestCase {
+
+    static let allTests = [
+        ("testSendsChildValues", testSendsChildValues),
+        ("testChildSubscribeDeadlock", testChildSubscribeDeadlock),
+        ("testCancelCancels", testCancelCancels),
+        ("testUpstreamDemandWithMaxPublishers", testUpstreamDemandWithMaxPublishers),
+        ("testUpstreamDemandWithNoMaxPublishers", testUpstreamDemandWithNoMaxPublishers),
+        ("testChildDemandWhenUnlimited", testChildDemandWhenUnlimited),
+        ("testChildDemandWhenLimited", testChildDemandWhenLimited),
+        ("testDemandFromLimitedtoUnlimited", testDemandFromLimitedToUnlimited),
+        ("testChildValueReceivedWhileSendingValue",
+         testChildValueReceivedWhileSendingValue),
+        ("testCompletesProperlyWhenChildrenOutliveUpstream",
+         testCompletesProperlyWhenChildrenOutliveUpstream),
+        ("testCompletesProperlyWhenUpstreamOutlivesChildren",
+         testCompletesProperlyWhenUpstreamOutlivesChildren),
+        ("testDoesNotCompleteWithBufferedValues", testDoesNotCompleteWithBufferedValues),
+        ("testFailsIfUpstreamFails", testFailsIfUpstreamFails),
+        ("testFailsIfChildFails", testFailsIfChildFails),
+        ("testFailsWithoutSendingBufferedValues", testFailsWithoutSendingBufferedValues),
+        ("testAllSubscriptionsReleasedOnUpstreamFailure",
+         testAllSubscriptionsReleasedOnUpstreamFailure),
+        ("testAllSubscriptionsReleasedOnChildFailure",
+         testAllSubscriptionsReleasedOnChildFailure),
+        ("testSendsSubcriptionDownstreamBeforeDemandUpstream",
+         testSendsSubcriptionDownstreamBeforeDemandUpstream),
+        ("testTestSuiteIncludesAllTests", testTestSuiteIncludesAllTests)
+    ]
+
+    func testSendsChildValues() {
+        let upstreamPublisher = PassthroughSubject<
+            PassthroughSubject<Int, TestingError>,
+            TestingError>()
+        let childPublisher1 = PassthroughSubject<Int, TestingError>()
+        let childPublisher2 = PassthroughSubject<Int, TestingError>()
+
+        let flatMap = upstreamPublisher.flatMap { $0 }
+
+        let downstreamSubscriber = TrackingSubscriber(receiveSubscription: {
+            $0.request(.unlimited)
+        })
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        upstreamPublisher.send(childPublisher1)
+        upstreamPublisher.send(childPublisher2)
+
+        childPublisher1.send(666)
+        childPublisher2.send(777)
+        XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
+                                                      .value(666),
+                                                      .value(777)])
+    }
+
+    // This test ensures that the code can properly re-enter when synchronously receiving
+    // a value during subscription (which Just(_) does).
+    //  1. FlatMap.Inner.receive(_ input:)
+    //    2. Publisher.subscribe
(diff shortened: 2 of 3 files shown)


Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        private func sendValueFromChild(_ value: NewPublisher.Output) {
+            serializedWork.enqueue { [weak self] in
+                _ = self?.downstream?.receive(value)
+            }
+        }
+    }
+}
+
```
OpenCombineBot: - :no_entry_sign: Files should have a single trailing newline. (`trailing_newline`)

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,175 @@
...
+    ///   - maxPublishers: The maximum number of publishers produced by this method.
+    ///   - transform: A closure that takes an element as a parameter and returns a
+    ///   publisher that produces elements of that type.
+    /// - Returns: A publisher that transforms elements from an upstream publisher into
+    /// a publisher of that element’s type.
+    public func flatMap<T, P>(maxPublishers: Subscribers.Demand = .unlimited,
+                              _ transform: @escaping (Self.Output) -> P)
+        -> Publishers.FlatMap<P, Self>
+        where T == P.Output, P : Publisher, Self.Failure == P.Failure {
+        return Publishers.FlatMap(upstream: self, maxPublishers: maxPublishers,
```
OpenCombineBot: - :no_entry_sign: Arguments should be either on the same line, or one per line. (`multiline_arguments`)

Sources/OpenCombine/Utilities/Utilities.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  Utilities.swift
+//
+//
+//  Created by Eric Patey on 17.08.2019.
+//
+
+/// A namespace for utility types.
+///
+public enum Utilities {}
```
broadwaylamb: We cannot introduce new public APIs to maintain source compatibilty with Combine, can you please change the access level to internal?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        func receive(completion: Subscribers.Completion<Failure>) {
+            // TODO: How does combine deal with failure? I presume I should do the same
+            // cleanup as in cancel after sending the completion.
+            switch completion {
+            case .finished:
+                decrementActiveSubscriptionCount()
+                break
+            case .failure(_):
```
OpenCombineBot: - :no_entry_sign: Arguments can be omitted when matching enums with associated types if they are not used. (`empty_enum_arguments`)

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        private func childCompleted(with completion: Subscribers.Completion<Failure>) {
+            switch completion {
+            case .finished:
+                // TODO: It may someday be useful to remove the subscription for this
+                // completed child
+                decrementActiveSubscriptionCount()
+                break
+            case .failure(_):
```
OpenCombineBot: - :no_entry_sign: Arguments can be omitted when matching enums with associated types if they are not used. (`empty_enum_arguments`)

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,177 @@
...
+            return Publishers.FlatMap(upstream: self,
+                                      maxPublishers: maxPublishers,
+                                      transform: transform)
+    }
+}
+
+extension Publishers {
+
+    public struct FlatMap<NewPublisher, Upstream>: Publisher
+    where NewPublisher: Publisher, Upstream : Publisher,
```
OpenCombineBot: - :no_entry_sign: Colons should be next to the identifier of the inheriting type (`inheritance_colon`)

epatey:
I'm now reasoning through how to do demand management for the child subscriptions. (demand for upstream was trivial since it's basically set to `maxPublishers` and never changed)

Experimenting with the Apple code, it seems to request 1 from all children, and then make some decisions about what to do next when it receives a value from any child.

I _think_ it will be necessarily complex to deal with buffering of one extra value per child subscription. Imagine the following flow:
- downstream requests initial demand of `.max(1)` of the flatMap
- upstream sends 3 values resulting in subscription to 3 children publishers
- each of those children is given a demand of `.max(1)`
- child[0] sends a value. flatMap passes it on to downstream which satisfies downstream's demand
- child[1] sends a value

At this point, flatMap has to buffer that value so that it can be passed to downstream if a later demand request is made.

continuing to experiment with Apple code, but I think it's unfortunately non-trivial

broadwaylamb:
@epatey hmm, that seems… unexpected. Yeah, FlatMap is probably the most nontrivial part of Combine.

epatey:
So, here's a unit test (that passes in apple combine mode), that helps explain more precisely the behavior. It's essentially that FlatMap puts `.max(1)` demand on all children until it has one buffered value ready to send downstream.

```swift
func testChildDemand() {
    let upstreamPublisher = PassthroughSubject<Void, Never>()

    var childDemand: Subscribers.Demand?
    let childSubscription = CustomSubscription(onRequest: { childDemand = $0 })
    let childPublisher = CustomPublisherBase<Int, Never>(
        subscription: childSubscription)

    let flatMap = upstreamPublisher.flatMap { _ in childPublisher }

    let downstreamSubscriber = TrackingSubscriberBase<Int, Never>(
        receiveSubscription: {
            downstreamSubscription = $0
            $0.request(.max(2))
    })

    flatMap.subscribe(downstreamSubscriber)

    upstreamPublisher.send(())

    // It seems that Apple creates enough demand on each child such that it can
    // have one extra/buffered value after the downstream demand is satisfied.
    // A demand of 1 on each child is created
    XCTAssertEqual(childDemand, .max(1))

    // Downstream demand is 2, so:
    //  - this value gets sent
    //  - downstream demand goes down to 1
    //  - child is asked for 1 more
    XCTAssertEqual(childPublisher.send(666), .max(1))
    XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
                                                  .value(666)])

    // Downstream demand is 1, so:
    //  - this value gets sent
    //  - downstream demand goes down to 0, but still need a buffered value
    //  - child is asked for 1 more
    XCTAssertEqual(childPublisher.send(777), .max(1))
    XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
                                                  .value(666),
                                                  .value(777)])

    // Downstream demand is 0, so:
    //  - this value is buffered and NOT sent
    //  - downstream demand is 0 and there's a buffered value
    //  - child is asked for 0 more
    XCTAssertEqual(childPublisher.send(888), .max(0))
    XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
                                                  .value(666),
                                                  .value(777)])

    childDemand = nil
    downstreamSubscription!.request(.max(2))
    XCTAssertEqual(childDemand, .max(1))
    XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
                                                  .value(666),
                                                  .value(777),
                                                  .value(888)])
}
```

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,280 @@
...
+            subscription: childSubscription)
+
+        let flatMap = upstreamPublisher.flatMap { _ in childPublisher }
+
+        var downstreamSubscription: Subscription?
+        let downstreamSubscriber = TrackingSubscriberBase<Int, Never>(
+            receiveSubscription: {
+                downstreamSubscription = $0
+                $0.request(.max(2))
+        })
```
OpenCombineBot: - :no_entry_sign: Closure end should have the same indentation as the line that started it. Expected 12, got 8. (`closure_end_indentation`)

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,280 @@
...
+        XCTAssertEqual(upstreamDemand, .max(2))
+    }
+
+    func testUpstreamDemandWithNoMaxPublishers() {
+        var upstreamDemand = Subscribers.Demand.max(666)
+        let upstreamSubscription = CustomSubscription(onRequest: { upstreamDemand = $0 })
+        let upstreamPublisher = CustomPublisherBase<Int, Never>(
+            subscription: upstreamSubscription)
+
+        let flatMap = upstreamPublisher.flatMap() { Just($0) }
```
OpenCombineBot: - :no_entry_sign: When using trailing closures, empty parentheses should be avoided after the method call. (`empty_parentheses_with_trailing_closure`)

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        /// Receive a new value from the upstream subscription. A new child subscription
+        /// will be made on the `NewPublisher` that the input value is transformed into.
+        /// - Parameter input: a value to be transformed by `transform`
+        func receive(_ input: Input) -> Subscribers.Demand {
+            let newChildPublisher = self.transform(input)
+            let newChildSubscriber = ChildSubscriber(
+                receiveValue: { [weak self] in
+                    self?.handleValue($1, fromChild: $0) ?? .none },
```
OpenCombineBot: - :no_entry_sign: Closure end should have the same indentation as the line that started it. Expected 16, got 66. (`closure_end_indentation`)

epatey:
Back pressure/demand makes things WAY WAY more complex. I really wonder if it was worth it vs Rx's decision to not have the feature. I've honestly never felt the need for it, but maybe I'm missing something. Getting close to complete support, though.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,381 @@
...
+                return
+            case .sendFinish:
+                downstream.receive(completion: .finished)
+                deactivate()
+                return
+            case .sendValues(let values):
+                var newDemand = Subscribers.Demand.none
+                values.forEach {
+                    newDemand += downstream.receive($0.value)
+                    // pausedChild is present only if the value was buffered and the child's
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 92 characters (`line_length`)

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,379 @@
...
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 92 characters (`line_length`)

epatey:
@broadwaylamb , I think I'm finally ready for a review. Here are a couple of thoughts:

- I factored out the code that deals with all of the complexity of the child subscriptions. By child subscriptions, I mean subscriptions of things that are not Upstream. This is where all of the complex demand, buffering logic is.
- I'm not in love with the name, ChildManagingInnerBase, but I wanted to start with a name that was precise. Feel free to propose something else.
- Although so far only FlatMap's Inner derives from that base class, I built it with Merge in mind. It's interesting that Merge is different than FlatMap because there is no Upstream for Merge.

epatey:
I'm thinking about iterating on `ChildManagingInnerBase` and recasting it as `MultiUpstreamBase`. In the case of `FlatMap`, I had been thinking about the upstream publisher and the child publishers. I think it would be a bit simpler if I just thought of them all as upstreams. Subclasses, i.e. `FlatMap.Inner`, could easily keep track of _special_ upstreams, but the base class really doesn't need to know or care about the distinction.

The challenge is that, in the case of `FlatMap` or `CombineLatest`, the different upstreams can be of completely different types. That's why I'm still just thinking about it. Suggestions are welcome.

epatey:
Ok, so my thinking has evolved a little. I think what I built with `ChildManagingInnerBase` is a class to help operators that deal with a dynamic list of multiple upstream publishers. By that I mean where the list is not fixed at compile time. It's possible that `FlatMap` is the only operator with that characteristic. `Merge` and `CombineLatest` have a fixed number of children.

If this proves to be correct, I'll undo the factoring I did since that base class adds complexity but no value.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,137 @@
...
+
+        /// The kind of values published by this publisher.
+        public typealias Output = Child.Output
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Upstream.Failure
+
+        public typealias Transform = (Upstream.Output) -> Child
```
broadwaylamb: Like I said, we can't add new public APIs.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+
+    internal final func deactivate() {
+        let foo = lock.do(_lockedDeactivate)
+        _unlockedDeactivate(foo)
+    }
+
+    internal final func evaluateActiveSubscriptionsExecuting(lockedWork: () -> Void) {
+        let shouldProcessQueue: Bool = lock.do {
+            lockedWork()
+            if childSubscribers.count == 0 && !lockedHasNonChildSubscription() {
```
broadwaylamb:
```suggestion
            if childSubscribers.isEmpty && !lockedHasNonChildSubscription() {
```

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+        let childrenToCancel = lock.do { () -> Set<ChildSubscriber> in
+            lockedCancel()
+            return _lockedDeactivate()
+        }
+        unlockedCancel()
+        _unlockedDeactivate(childrenToCancel)
+    }
+
+    internal final func deactivate() {
+        let foo = lock.do(_lockedDeactivate)
```
broadwaylamb: `foo`? :)

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    private func demandForChild() -> Subscribers.Demand {
+        return self.downstreamDemand == .unlimited ? .unlimited : .max(1)
+    }
+
+    private func child(_ child: ChildSubscriber,
+                       receivedValue value: Output) -> Subscribers.Demand {
+        // Obtain the lock in order to determine what to do
+        let (valueWasBuffered, processTheQueue): (Bool, Bool) = lock.do {
```
broadwaylamb: Let's either name this `drainTheQueue`, or name the return values below `processTheQueue `, just to be consistent.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    internal final func sendFailureDownstream(failure: Failure) {
+        downstream?.receive(completion: .failure(failure))
+        deactivate()
+    }
+
+    internal final func addChild(_ newChild: Child) {
+        let newChildSubscriber = ChildSubscriber(
+            receiveValue: { [weak self] in
```
broadwaylamb: As far as I understand, `ChildSubscriber`s should not be able to outlive `self`, hence, it probably makes sense to use `unowned` here.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    private enum QueueWorkStatus {
+        case noWork
+        case sendFinish
+        case sendValues(values: [WorkType])
+    }
+
+    private func processQueue() {
+        assert(processingQueue)
+        let work: QueueWorkStatus = lock.do {
+            if downstreamDemand == .none || valuesToSend.count == 0 {
```
broadwaylamb:
```suggestion
            if downstreamDemand == .none || valuesToSend.isEmpty {
```

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+                if sendFinishedAfterDrainingQueue && valuesToSend.count == 0 {
```
broadwaylamb:
```suggestion
                if sendFinishedAfterDrainingQueue && isEmpty {
```

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    private func removeActiveSubscription(forChild child: ChildSubscriber) {
+        evaluateActiveSubscriptionsExecuting { childSubscribers.remove(child) }
+    }
+
+    private enum QueueWorkStatus {
+        case noWork
+        case sendFinish
+        case sendValues(values: [WorkType])
```
broadwaylamb: Let's use `ArraySlice<WorkType>` here.
```suggestion
        case sendValues(values: ArraySlice<WorkType>)
```

This will allow us to avoid allocating a whole array below.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+                    return .sendFinish
+                } else {
+                    processingQueue = false
+                    return .noWork
+                }
+            }
+
+            let countToSend = Int(min(UInt(valuesToSend.count),
+                                      downstreamDemand.rawValue))
+            let result = Array(valuesToSend[0..<countToSend])
```
broadwaylamb:
```suggestion
            let result = valuesToSend[..<countToSend]
```

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+                if sendFinishedAfterDrainingQueue && valuesToSend.count == 0 {
+                    return .sendFinish
+                } else {
+                    processingQueue = false
+                    return .noWork
+                }
+            }
+
+            let countToSend = Int(min(UInt(valuesToSend.count),
+                                      downstreamDemand.rawValue))
```
broadwaylamb: We can avoid integer conversion here:

```swift
let countToSend = min(valuesToSend.count, downstreamDemand.max ?? .max)
```

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+            }
+
+            if newDemand != .none {
+                lock.do { downstreamDemand += newDemand }
+            }
+        }
+
+        // Loop to try again in case somebody put stuff on the queue while we were
+        // sending values with the lock unlocked.
+        processQueue()
```
broadwaylamb: Can we just use a while loop instead of recursion?

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+            processingQueue = true
+            return (true, demandModeChanged)
+        }
+
+        if demandModeChanged {
+            // TODO: This code is currently asymmetric because Apple only updates the
+            // child demand when going from limited to unlimited. It does not update the
+            // demand when goinf from unlimited to limited (which seems like a bug to me).
+            if demand == .unlimited {
+                // TODO: This code isn't yet thread safe. The correct change is to do this
```
broadwaylamb: Is this TODO still relevant?

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    internal func request(_ demand: Subscribers.Demand) {
+        let (drainTheQueue, demandModeChanged) = lock.do { () -> (Bool, Bool) in
+            let demandModeChanged =
+                (demand == .unlimited) != (downstreamDemand == .unlimited)
+            downstreamDemand = demand
+            if processingQueue {
+                return (false, demandModeChanged)
+            }
+            processingQueue = true
+            return (true, demandModeChanged)
```
broadwaylamb: It seems like we can replace this with a simpler code:

```swift
let demandModeChanged =
    (demand == .unlimited) != (downstreamDemand == .unlimited)
downstreamDemand = demand
defer { processingQueue = true }
return (!processingQueue, demandModeChanged)
```

Otherwise it's not immediately clear that `processingQueue` is always `true` at the end of this closure.

Also, can you please rename it to `isProccessingQueue` so that the valriable reads like a bool, not some queue object?

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    ///   child's demand reached .none - effectively pausing the child.)
+    internal final class ChildSubscriber: Hashable {
+        internal typealias Input = Downstream.Input
+        internal typealias Failure = Downstream.Failure
+
+        private var _upstreamSubscription: Subscription?
+        private let _receiveValue: (ChildSubscriber, Input) -> Subscribers.Demand
+        private let _receiveCompletion: (ChildSubscriber, Subscribers.Completion<Failure>)
+        -> Void
+        private let _initialDemandForChild: () -> Subscribers.Demand
```
broadwaylamb: I'm not sure how it plays with your plans regarding `Merge`, but I'm curious why are you using closures here? I can see that they're only set in the `addChild(_:)` method, but the logic in these closures is always the same, which means that they could probably be replaced with plain methods, and the `ChildSubscriber` could just store an unowned reference to `InnerWithChildrenBase`.

Storing closures can impose certain overhead.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+extension InnerWithChildrenBase.ChildSubscriber: Subscriber {
+    internal func receive(subscription: Subscription) {
+        if _upstreamSubscription == nil {
+            _upstreamSubscription = subscription
+            subscription.request(_initialDemandForChild())
+        } else {
+            assert(false)
```
broadwaylamb: There's `assertionFailure()`.

Maybe you can try to trigger this code path in Combine and see what it does? If it's not possible, than I'd maybe even go with `fatalError("unreachable")`.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+
+        childPublisher1.send(666)
+        childPublisher2.send(777)
+        XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
+                                                      .value(666),
+                                                      .value(777)])
+    }
+
+    // This test ensures that the code can properly re-enter when synchronously receiving
+    // a value during subscription (which Publishers.Just does).
```
broadwaylamb: nit: there's no `Publishers.Just` anymore, just `Just` :)

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        })
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        XCTAssertEqual(upstreamPublisher.send(1), .none)
+
+        downstreamSubscription?.cancel()
+
+        XCTAssertEqual(upstreamSubscription.history.last!, .cancelled)
+        XCTAssertEqual(childSubscription.history.last!, .cancelled)
```
broadwaylamb: `Optional` conditionally conforms to `Equatable`, so there's no need to force-unwrap here. We don't want to crash in tests, we want to fail gracefully.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    func testUpstreamDemandWithMaxPublishers() {
+        var upstreamDemand = Subscribers.Demand.unlimited
```
broadwaylamb: Let's make it an optional with no initial value, so that the intent is clear.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        let downstreamSubscriber = TrackingSubscriberBase<Int, Never>(
+            receiveSubscription: { $0.request(.unlimited) })
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        XCTAssertEqual(upstreamDemand, .max(2))
+    }
+
+    func testUpstreamDemandWithNoMaxPublishers() {
+        var upstreamDemand = Subscribers.Demand.max(666)
```
broadwaylamb: Same here, why use magic constants when we have optionals? :)

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        var downstreamSubscription: Subscription?
+        let downstreamSubscriber = TrackingSubscriberBase<Int, Never>(receiveSubscription:
+        {
+            downstreamSubscription = $0
+            $0.request(.max(2))
+        })
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        upstreamPublisher.send(())
```
broadwaylamb: Can you please use this overload here and below?
```suggestion
        upstreamPublisher.send()
```

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        //  - this value is buffered and NOT sent
+        //  - downstream demand is 0 and there's a buffered value
+        //  - child is asked for 0 more
+        XCTAssertEqual(childPublisher.send(888), .max(0))
+        XCTAssertEqual(downstreamSubscriber.history, [.subscription("FlatMap"),
+                                                      .value(666),
+                                                      .value(777)])
+
+        childDemand = nil
+        downstreamSubscription!.request(.max(2))
```
broadwaylamb:
```suggestion
        try XCTUnwrap(downstreamSubscription).request(.max(2))
```

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        flatMap.subscribe(downstreamSubscriber)
+
+        upstreamPublisher.send(())
+
+        XCTAssertEqual(childDemand, .unlimited)
+
+        downstreamSubscription?.request(.max(3))
+        // TODO: Strangely Apple doesn't reset the child demand down to 1. That totally
+        // seems like a bug
+//        XCTAssertEqual(childDemand, .max(1))
```
broadwaylamb: I don't think this is a bug.

As we've discovered, demand is additive, which means that if you request `.unlimited` and after that you request `.max(42)`, it should be completely the same as just requesting `.unlimited` once. If you blow up a dam, you can't stop the flow amymore :)

I believe this is what's happening here.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+
+        let childSubscription = CustomSubscription()
+        let child = CustomPublisher(subscription: childSubscription)
+
+        let flatMap = upstreamPublisher.flatMap { _ in child }
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        XCTAssertEqual(upstreamPublisher.send(1), .none)
+        _ = child.send(666) // TODO: demand
```
broadwaylamb: Let's wrap this in `XCTAssertEqual` instead of ignoring the value.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+
+        let children = [AnyPublisher(child1), AnyPublisher(child2)]
+        let flatMap = upstreamPublisher.flatMap { children[$0] }
+
+        flatMap.subscribe(downstreamSubscriber)
+
+        XCTAssertEqual(upstreamPublisher.send(0), .none)
+        XCTAssertEqual(upstreamPublisher.send(1), .none)
+        _ = child1.send(666)    // TODO: demand
+        _ = child2.send(777)    // TODO: demand
```
broadwaylamb: And here

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
```diff
@@ -0,0 +1,539 @@
...
+        child1.send(completion: .failure(TestingError.oops))
```
broadwaylamb, line 587: Doesn't it violate the contract? AFAIK, the upstream should wait for all the current children to finish before sending the next value. Also, `child2` should not send any values before `child1` is finished.

If it is intentional, can you add a comment?

broadwaylamb (changes requested):
Thank you for your hard work, Eric!
You've done a really amazing job. I like that you've put a lot of comments, since the implementation is nontrivial indeed. I also like the tests. I have some suggestions and questions though.

I'd like to ask you to not force-push for now as it makes reviewing a little bit harder. After it's complete, you can squash the commits to keep the history clear.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Ooops. Forgot about that...cleaning that all up. Also, gonna just stick with the names used by Apple's interface.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
epatey: Are you comfortable turning on that rule in SwiftLint? It could have caught it.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
epatey: If you feel strongly, I'll make that change.

In general, though, I feel that it's just not worth that sort of analysis. The downsides of getting it wrong are huge, and the upside of using `unowned` is very minor. It just doesn't seem worth it to me...unless, we see data suggesting that the `weak` operations are material from a performance perspective.

Put a different way, whenever I see `unowned` while I'm doing a code review, it's a big flag and I have to spend a big amount of energy proving to myself that it's safe.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
broadwaylamb: Hm, I was pretty sure it was turned on, but now I see it isn't. Can you please turn it on here?

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
epatey: I _think_ it is still relevant, but it's an extremely edge case. It opens a potential timing window when switching between `.unlimited` demand and `.max` demand. As suggested by the TODO just above, Apple themselves have bugs in this area. My thinking was that I wanted to get this PR merged and then do another small refactoring to deal with that edge case.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
epatey: Yeah. My recent realization about makes this all moot. I'll be getting rid of the base class since I actually think that there are no other consumers of it.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
broadwaylamb: It just creates dead code that will never be executed, which I find misleading.
I also believe that if invariants are broken, we should not propagate magic constants instead of telling that something went wrong.

If `ChildSubscriber` ever outlives `self` (maybe due to a bug that will be introduced later), it is our fault, not the client's, and we have better chances to spot it if there's a crash, and the client will not have to spend hours figuring out why something silently behaves wrong.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
epatey: Yeah. This makes total sense. I didn't realize that I could do an equality test against an optional. Cool.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
epatey: So you're saying that even when calling `.request` on the subscription, that it's additive? I had been thinking of the additive behavior only for the result of `.receive(_ input:)` If so, I see. I'll make another pass.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
epatey: I missed this one, and I've made the change.

I would suggest, though, that reasonable people could differ on asserting on that line everywhere it occurs in every test. The other perspective is that a test should be testing one and only one thing. That is, the set of asserts made within a test should really be in support of what that test's name says the test is about. Instead, there could be a variety of tests specifically around confirming the result of child subscriptions sending the values to the FlatMap.

I don't feel strongly about it, but thought I'd just share a different world view.

Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift
epatey: I was curious about the same thing, and that's why I wrote the test - in order to cement what Apple's behavior is. In a nutshell, I think this test proves that failures "jump the queue". This means that if there's a failure, pending/buffered values will NOT be sent. The failure causes the whole thing to fail immediately.

Sources/OpenCombine/Publishers/ChildManagingInnerBase.swift
```diff
@@ -0,0 +1,324 @@
...
+        case sendFinish
+        case sendValues(values: ArraySlice<WorkType>)
+    }
+
+    private func processQueue() {
+        assert(queueIsBeingProcessed)
+
+        // We loop processing the queue in case somebody put stuff on the queue while we
+        // were sending values with the lock unlocked.
+        while (true) {
```
OpenCombineBot: - :no_entry_sign: `if`, `for`, `guard`, `switch`, `while`, and `catch` statements shouldn't unnecessarily wrap their conditionals or arguments in parentheses. (`control_statement`)

epatey:
@broadwaylamb, I've gotten rid of the base class in my last commit. As I feared, there really are no other operators that have to manage a _dynamic_ set of child subscriptions. Unfortunately, that commit moved a whole lot of code, so you may want to review the previous commits individually before looking at that big one to make understanding my reaction to your feedback easier.

epatey:
@broadwaylamb, let me know if/when you'd like me to rebase, squash and merge this - or if you'd like additional changes.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+}
+
+extension Publishers.FlatMap {
+    fileprivate final class Inner<Downstream: Subscriber>
+        : CustomStringConvertible, Cancellable
+    where Downstream.Input == Child.Output, Downstream.Failure == Upstream.Failure {
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        // TODO: Implement CustomReflectable. See implementation in OperatorSubscription.
```
broadwaylamb: Let's do it

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    private func cancelChildren(_ childrenToCancel: Set<ChildSubscriber>) {
+        childrenToCancel.forEach { $0.cancel() }
+    }
+
+    private final func evaluateActiveSubscriptionsExecuting(lockedWork: () -> Void) {
+        let shouldProcessQueue: Bool = lock.do {
+            lockedWork()
+            if childSubscribers.isEmpty && upstreamSubscription == nil {
```
broadwaylamb, line 157: Can you please add a comment describing what this code is doing and what its purpose is?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+    private func removeActiveSubscription(forChild child: ChildSubscriber) {
+        evaluateActiveSubscriptionsExecuting { childSubscribers.remove(child) }
+    }
+
+    private func child(_ child: ChildSubscriber,
+                       receivedValue value: Child.Output) -> Subscribers.Demand {
+        // Obtain the lock in order to determine what to do
+        let (valueWasBuffered, processTheQueue): (Bool, Bool) = lock.do {
+            // Consider pending values to send when assessing downstreamDemand
+            if downstreamDemand - valuesToSend.count == .none {
```
broadwaylamb: I can't immediately tell if it's possible for `downstreamDemand` to be less than `valuesToSend.count`, but if it **is** possible, this branch will be entered due to how subtraction for `Demand` works: `.max(1) - .max(3)` will return `.none`.

This code is equivalent to `downstreamDemand <= valuesToSend.count`. If it is intentional, let's add a comment and use the `<=` overload. If not, i suggest to use the `==` overload to compare the values for equality.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+
+    /// Receive a new value from the upstream subscription. A new child subscription
+    /// will be made on the `Child` that the input value is transformed into.
+    /// - Parameter input: a value to be transformed by `transform`
+    fileprivate func receive(_ input: Input) -> Subscribers.Demand {
+        let newChildSubscriber = ChildSubscriber(
+            receiveValue: { [unowned self] in self.child($0, receivedValue: $1) },
+            receiveCompletion: { [unowned self] in self.child($0, completedWith: $1) },
+            initialDemandForChild: { [unowned self] in self.demandForChild() }
+        )
```
broadwaylamb: Like I said, instead of passing closures let's initialize `ChildSubscriber` with reference to `self`:

```swift
let newChildSubscriber = ChildSubscriber(self)
```

What I mean is that `ChildSubscriber` has a property `unowned var enclosing: Inner` and its methods just use that property instead of calling the closures:

```swift
extension Publishers.FlatMap.Inner.ChildSubscriber: Subscriber {
    internal func receive(subscription: Subscription) {
        if _upstreamSubscription == nil {
            _upstreamSubscription = subscription
            subscription.request(enclosing.demandForChild())
        } else {
            assertionFailure()
            subscription.cancel()
        }
    }

    fileprivate func receive(_ input: Input) -> Subscribers.Demand {
        return enclosing.child(self, receivedValue: input)
    }

    fileprivate func receive(completion: Subscribers.Completion<Failure>) {
        return enclosing.child(self, completedWith: completion)
    }
}
```

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+        : CustomStringConvertible, Cancellable
+    where Downstream.Input == Child.Output, Downstream.Failure == Upstream.Failure {
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        // TODO: Implement CustomReflectable. See implementation in OperatorSubscription.
+
+        private typealias WorkType = (
+            value: Downstream.Input,
+            pausedChild: ChildSubscriber?)
```
broadwaylamb, line 88: Can you please document here when `pausedChild` is `nil`? And maybe rename this typealias to `PendingValue` or something like that.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    private func removeActiveSubscription(forChild child: ChildSubscriber) {
+        evaluateActiveSubscriptionsExecuting { childSubscribers.remove(child) }
+    }
+
+    private func child(_ child: ChildSubscriber,
+                       receivedValue value: Child.Output) -> Subscribers.Demand {
+        // Obtain the lock in order to determine what to do
+        let (valueWasBuffered, processTheQueue): (Bool, Bool) = lock.do {
```
broadwaylamb: I'm having trouble understanding the logic in this block, can you add a comment?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+
+    private func demandForChild() -> Subscribers.Demand {
+        return self.downstreamDemand == .unlimited ? .unlimited : .max(1)
+    }
+
+    private enum QueueWorkStatus {
+        case noWork
+        case sendFinish
+        case sendValues(values: ArraySlice<WorkType>)
+    }
```
broadwaylamb, line 232: Maybe move this enum into the `processQueue` function?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+                        return .sendFinish
+                    } else {
+                        queueIsBeingProcessed = false
+                        return .noWork
+                    }
+                }
+
+                let countToSend = min(valuesToSend.count, downstreamDemand.max ?? .max)
+                let result = valuesToSend[0..<countToSend]
+                valuesToSend.removeFirst(countToSend)
```
broadwaylamb, line 253: This operation is O(n). For now let's leave it as is, but in the future we may need to use a better data structure than array here.

Can you add a `TODO`?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
```diff
@@ -0,0 +1,402 @@
...
+            case .sendValues(let values):
+                var newDemand = Subscribers.Demand.none
+                values.forEach {
+                    newDemand += downstream.receive($0.value)
+                    // pausedChild is present only if the value was buffered and the
+                    // child's demand was left at `.none`. In that case, once we send the
+                    // buffered value, we need to tell the child to get another value.
+                    if let pausedChild = $0.pausedChild {
+                        pausedChild.request(.max(1))
+                    }
```
broadwaylamb: This `if` can be replaced with `$0.pausedChild?.request(.max(1))`

broadwaylamb (changes requested):

broadwaylamb:
`XCTestManifests.swift` is not needed anymore, can you synchronize with master and remove this file?

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Good catch. I think, as currently written, the code will fail to send any values if, including this new value, we'll have buffered more than the downstream demand. That's not the goal. The goal would be to send up to the downstream demand.

Of course, Apple could do something strange, so I'll have to add a few more test scenarios. I'll add some more unit tests around this.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Yup. I get what you mean now. I had so many balls in the air at the time you said that last time I missed it. In fact, I've done it the way you want in the `.zip` implementation. I'll try to iterate on that tomorrow.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Yeah. This is the same code from above. I'll add more documentation after I fix the potential bug you uncovered.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Good idea, but that causes a compile error followed by an llvm crash. Lol.

error: type 'QueueWorkStatus' cannot be nested in generic function 'processQueue()'
        enum QueueWorkStatus {

broadwaylamb: Oops! I think Apple folks would appreciate a bug report on [bugs.swift.org](https://bugs.swift.org).

broadwaylamb: But yeah, my bad, I forgot that nested types are not allowed in generic contexts.

Sources/OpenCombine/Publishers/Publishers.FlatMap.swift
epatey: Sure. I have to tell you, though, that I'm kind of shocked at the inconsistency of how Apple has been dealing with stuff related to demand. I may be missing something, but their zip implementation, for example seems completely bogus. Zip is pretty much the canonical example of the need for backpressure. Imagine zipping two sources. One fires constantly, and the other never. By setting a demand of 1 to each child, zip could avoid having to buffer a potentially infinite number of values from the chatty upstream. But Apple didn't bother to do that. Super strange.

epatey:
@broadwaylamb, it looks like the swift 5.0 build for this pr is failing, but the only thing I see in the logs [here](https://travis-ci.org/broadwaylamb/OpenCombine/jobs/584259390) is a problem with the swiftlint configuration. I'm guessing that's the reason it's failing, but I'm not sure. Any thoughts?

broadwaylamb:
@epatey oh. My bad, I removed `XCTestManifests.swift`, but forgot to remove the Danger rule for that file. Should be fine now.

broadwaylamb (approved):
Thank you so much for this PR!

PR #49, opened 2019-08-28
Implement Scan/TryScan and convert Map/TryMap to use common base class.

epatey:
This `Scan`/`TryScan` implementation shares a common `TransformingInner` class with `Map` and `TryMap`.


Tests/OpenCombineTests/PublisherTests/ScanTests.swift
```diff
@@ -0,0 +1,347 @@
...
+            .completion(.failure(expectedError)),
+            .completion(.failure(expectedError))
+        ])
+    }
+
+    func testTryScanFailureBecauseOfThrow() {
+        var counter = 0 // How many times the transform is called?
+
+        let publisher = PassthroughSubject<Int, Error>()
+        let scan = publisher.tryScan(0) { (accum, value) -> Int in
```
OpenCombineBot: - :no_entry_sign: Parentheses are not needed when declaring closure arguments. (`unneeded_parentheses_in_closure_argument`)
- :no_entry_sign: Unused parameter "accum" in a closure should be replaced with _. (`unused_closure_parameter`)

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  TransformingInner.swift
+//
+//  Created by Eric Patey on 26.08.2019.
+//
+
+internal class TransformingInner<Upstream: Publisher, Downstream: Subscriber>
```
broadwaylamb, line 16: To be honest, I'm not in favor of this `TransformingInner` class, sorry.

It seems to use stored properties like `shouldProxySubscription` and `description` just to mimic Combine, while not adding any value beyond that, but possibly introducing overhead.

My suggestion would be to leave `Map` and `TryMap` as is, and implement `Scan` and `TryScan` using an ad-hoc subscription class. Yep, it's a little bit of duplication, but at least it doesn't make things more complex.

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,89 @@
...
+    : TransformingInnerBase<Upstream, Downstream>,
+    Subscriber
+    where Upstream.Failure == Downstream.Failure
+{
+    internal init(downstream: Downstream, transform: @escaping (Input)
+        -> Downstream.Input) {
+        super.init(downstream: downstream, transform: catching(transform))
+    }
+
```
OpenCombineBot: - :no_entry_sign: Ensure definitions have a lower access control level than their enclosing parent (`lower_acl_than_parent`)

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,89 @@
...
+    Subscriber
+    where Upstream.Failure == Downstream.Failure,
+    Downstream.Failure == Error
+{
+    internal init(downstream: Downstream, transform: @escaping (Input) throws
+        -> Downstream.Input) {
+        super.init(downstream: downstream, transform: catching(transform))
+    }
+
+    public final func receive(subscription: Subscription) {
```
OpenCombineBot: - :no_entry_sign: Ensure definitions have a lower access control level than their enclosing parent (`lower_acl_than_parent`)

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,89 @@
...
+    : TransformingInnerBase<Upstream, Downstream>,
+    Subscriber
+    where Upstream.Failure == Downstream.Failure
+{
+    internal init(downstream: Downstream, transform: @escaping (Input)
+        -> Downstream.Input) {
+        super.init(downstream: downstream, transform: catching(transform))
+    }
+
+    public final func receive(subscription: Subscription) {
```
OpenCombineBot: - :no_entry_sign: Ensure definitions have a lower access control level than their enclosing parent (`lower_acl_than_parent`)

Sources/OpenCombine/Publishers/TransformingInner.swift
epatey: @broadwaylamb, thanks for the review and feedback. I appreciate it.

I'd like to discuss it a little bit further to make sure I'm aligned specifically with the sorts of things that you consider unnecessary complexity.

I think it's worth the extra effort, because I'm motivated very strongly by code sharing and a D.R.Y. approach. If done properly, I think sharing as much code as possible between operators will have massive benefit and allow us to deliver a more consistently correct implementation.

That said, I agree that some of my shared implementation was a bit lazy. Specifically, some of my use of stored properties as you pointed out. I think the generalized concern is paying any memory/runtime expense to support `.tryScan` if I only ever use `.scan`.

My use of `shouldProxySubscription` was a great example of failing that test. Though it didn't incur a memory penalty, my implementation of `description` in the base class did cause unneeded runtime cost during init time.

I've made another pass as the shared class to get rid of all of those stored properties and made it simpler to consume. It now deals with all of the following differences between the throwing and non-throwing variants without leaking any cost/complexity between the two:
- Conforming to `Subscription` in the throwing variant and passing `self` 
- Differences in the implementation of `Subscriber.receive(subscription:)` to pass the proper `Subscription` downstream: the upstream subscription for non-throwing variant and `self` for the throwing variant. 
- Adapting `Subscribers.Completion<Upstream.Failure>` to `Subscribers.Completion<Error>` for the throwing variant.

The more I work on this stuff, the more I am convinced that there are many non-obvious Apple behaviors that we need to match. Because of this, I think it would be a mistake to have similar but copied and likely divergent code across the project. Here's one example that we're already soaking in. Apple's behavior if `.receive(completion:)` is called multiple times is non-obvious. They allow multiple `.failure`'s to go downstream, but they do NOT allow multiple `.finished`'s to go downstream. It's not crazy complex, but it's subtle and non-obvious. As currently implemented, this is tested properly for `.map`, but not for `.tryMap`. In fact, the code is broken for `.tryMap`. Simply add a copy of `testError` that uses `tryMap` rather than `map`. My point is that we already have inconsistencies that have caused bugs. Properly tested shared code would benefit ALL consumers of that shared code.

So, that all said, I'd really appreciate it if you took another pass a the code now that I've incorporated your feedback.

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,97 @@
...
+// manages `Downstream` and `Upstream` references.
+//
+// Operator implementations implement `Inner` classes that derive from one of the two
+// internal subclasses of this base class - `NonThrowingTransformingInner` or
+// `ThrowingTransformingInner`.
+
+internal class TransformingInner<Upstream: Publisher, Downstream: Subscriber>
+    : OperatorSubscription<Downstream>
+{
+    private final var transform: ((Upstream.Output) -> Result<Downstream.Input, Downstream.Failure>)?
```
OpenCombineBot, line 19: - :no_entry_sign: Line should be 90 characters or less: currently 101 characters (`line_length`)

broadwaylamb:
Can you please synchronize with master and fix the conflicts?

Sources/OpenCombine/Publishers/TransformingInner.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+internal class ThrowingTransformingInner<Upstream: Publisher, Downstream: Subscriber>
+    : TransformingInner<Upstream, Downstream>,
+    Subscriber,
+    Subscription
+    where Downstream.Failure == Error
+{
+    typealias Failure = Upstream.Failure
```
OpenCombineBot, line 76: - :no_entry_sign: All declarations should specify Access Control Level keywords explicitly. (`explicit_acl`)

broadwaylamb:
I'm sorry it took so long to land this. I've been working on a more efficient architecture and also reverse-engineering Combine to see how it does things in the implementations of its operators.

I've cherry-picked this in #83. I had to remove the common base class, because it turns out that `Map`'s `Inner` in Combine is not a class at all, it's a struct! It makes sense though, because that `Inner` doesn't have any state. `Scan`, in contrast, does have a state. In addition, the implementations of `Scan` and `TryScan` don't use common logic because they're quite different: `Scan` just forwards the calls to its `Subscriber` methods to the downstream subscriber, and `TryScan` maintains internal subscription state.

I want to thank you for the work you've done, that's been really a huge help!

PR #52, opened 2019-09-01
@propertyWrapper Published

beastgrim:
Hello! I've implemented this interface. That allows use construction like this: 
` @Published var value: String?`
And bind and subscribe:
`$value.assign(to: \.text, on: textLabel)`
I hope that will be helpful.


Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,65 @@
...
+        
+        // MARK: - Private
+        private var _passthrowObject = OpenCombine.PassthroughSubject<Value,Never>()
+        
+        internal func send(_ input: Output) {
+            _passthrowObject.send(input)
+        }
+    }
+
+    /// The property that can be accessed with the `$` syntax and allows access to the `Publisher`
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 98 characters (`line_length`)

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,65 @@
...
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<S>(subscriber: S) where Value == S.Input, S : Subscriber, S.Failure == Published<Value>.Publisher.Failure {
+            
+            _passthrowObject.receive(subscriber: subscriber)
+        }
+        
+        
+        // MARK: - Private
+        private var _passthrowObject = OpenCombine.PassthroughSubject<Value,Never>()
```
OpenCombineBot: - :no_entry_sign: There should be no space before and one after any comma. (`comma`)

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  Published.swift
+//  OpenCombine
+//
+//  Created by Евгений Богомолов on 01/09/2019.
+//
+
+import Foundation
+
+@propertyWrapper public struct Published<Value> {
```
broadwaylamb, line 15: Let's wrap this whole file in conditional compilation block to be compatible with Swift 5:

```swift
#if swift(>=5.1)
...
#endif
```

Sources/OpenCombine/Published.swift
broadwaylamb: Nit: `_passthroughSubject`. Probably should be `let` instead of `var`.

Also, this `MARK` probably doesn't add much new information since this is the only private property in this type.

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,65 @@
...
+
+        /// This function is called to attach the specified `Subscriber` to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<S>(subscriber: S) where Value == S.Input, S : Subscriber, S.Failure == Published<Value>.Publisher.Failure {
+            
+            _passthrowObject.receive(subscriber: subscriber)
```
broadwaylamb: It is [discouraged](https://developer.apple.com/documentation/combine/publisher/3204756-subscribe) to call this method, can you please change it to `_passthroughSubject.subscribe(subscriber)`?

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,65 @@
...
+    
+    public var wrappedValue: Value {
+        get { value }
+        set {
+            value = newValue
+            projectedValue.send(newValue)
+        }
+     }
+    
+    // MARK: - Private
```
broadwaylamb: There's a single private member in this type, probably not worth a `MARK`.

broadwaylamb (changes requested):
Thank you!

I've added some comments, but in order for it to be merged we'll need tests, also can you please address the bot's comments? :)

beastgrim:
Sorry for inconvenience. It is first time when I do pull request to GitHub. I fixed all comments and also warnings from SwiftLint. It remains only to write tests. 
When it is done what I should do next? Should I close this pull request and open another or there is another variants? Thanks!

Sources/OpenCombine/Published.swift
beastgrim: Done

broadwaylamb:
>It is first time when I do pull request to GitHub

I’m flattered!

>Should I close this pull request and open another or there is another variants?

When you write the tests, commit them to this branch, it is better if they come along with the feature that they test. No need to create a separate pull request for that :)

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,70 @@
...
+        ///                   once attached it can begin to receive values.
+        public func receive<SubscriberType>(subscriber: SubscriberType)
+            where Value == SubscriberType.Input,
+            SubscriberType: Subscriber,
+            SubscriberType.Failure == Published<Value>.Publisher.Failure
+        {
+            _passthrowObject.subscribe(subscriber)
+        }
+
+        private let _passthrowObject = OpenCombine.PassthroughSubject<Value, Never>()
```
broadwaylamb: I've just checked, Combine uses `CurrentValueSubject` here.

I've run the following code:

```swift
import Combine

var p = Published(initialValue: 42)
dump(p.projectedValue, maxDepth: 1)
```
And got:

```
▿ Combine.Published<Swift.Int>.Publisher
  ▹ subject: Combine.CurrentValueSubject<Swift.Int, Swift.Never> #0
```

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,70 @@
...
+        internal func send(_ input: Output) {
+            _passthrowObject.send(input)
+        }
+    }
+
+    /// The property that can be accessed with the
+    /// `$` syntax and allows access to the `Publisher`
+    public private(set) lazy var projectedValue: Published<Value>.Publisher = .init()
+
+    public var wrappedValue: Value {
```
broadwaylamb, line 76: [This property is marked `unavailable` in Combine, we should do this too](https://gist.github.com/broadwaylamb/82dc2ce4ffbe06527c2c352b8f10910f#file-combine-swiftinterface-swift-L812-L816)

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,70 @@
...
+            projectedValue.send(newValue)
+        }
+     }
+
+    private var _value: Value
+
+    /// For removing warning: Property wrapper's `init(initialValue:)`
+    /// should be renamed to 'init(wrappedValue:)';
+    /// use of 'init(initialValue:)' is deprecated
+    public init(wrappedValue: Value) {
```
broadwaylamb: Since this is not part of public interface of Combine, we should not have this initializer. But we have a warning, yeah.

I propose the following: let's mark is as `unavailable` too:

```swift
@available(*, unavailable)
public init(wrappedValue: Value) {
    self._value = wrappedValue
}
```

Now there is no warning, and we don't diverge from the Combine's public API, because clients won't be able to call this initializer. When/if Combine adds this initializer to its interface, we can easily remove this availability attribute.

broadwaylamb (changes requested):
Can you please remove the interface of `Published` from the [RemainingCombineInterface.swift](https://github.com/broadwaylamb/OpenCombine/blob/8b25238154056584aaccf57f5231e5d32b041931/RemainingCombineInterface.swift#L3215-L3247) so that we can track what's implemented and what's not?

I also see that your implementation lacks [this subscript](https://github.com/broadwaylamb/OpenCombine/blob/8b25238154056584aaccf57f5231e5d32b041931/RemainingCombineInterface.swift#L3224-L3228). Implementing it is non-trivial, so I suggest to add a stub marked as unavailable:

```swift
@available(*, unavailable, message: "This subscript is unavailable in OpenCombine yet")
public static subscript<EnclosingSelf: AnyObject>(
    _enclosingInstance object: EnclosingSelf,
    wrapped wrappedKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Value>,
    storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Published<Value>>
) -> Value {
    get { fatalError() }
    set { fatalError() }
}
```

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,70 @@
...
+        /// `Subscriber` to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<SubscriberType>(subscriber: SubscriberType)
+            where Value == SubscriberType.Input,
+            SubscriberType: Subscriber,
+            SubscriberType.Failure == Published<Value>.Publisher.Failure
```
broadwaylamb: Can you also please make this long method signature a bit shorter, like this:

```swift
public func receive<Downstream: Subscriber>(subscriber: Downstream)
    where Downstream.Input == Value, Downstream.Failure == Never
```

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,70 @@
...
+        {
+            _passthrowObject.subscribe(subscriber)
+        }
+
+        private let _passthrowObject = OpenCombine.PassthroughSubject<Value, Never>()
+
+        internal func send(_ input: Output) {
```
broadwaylamb: Instead of adding this new method you can just mark the subject `fileprivate` and use it directly from `Published`

beastgrim (commented):
> I also see that your implementation lacks [this subscript](https://github.com/broadwaylamb/OpenCombine/blob/8b25238154056584aaccf57f5231e5d32b041931/RemainingCombineInterface.swift#L3224-L3228). Implementing it is non-trivial, so I suggest to add a stub marked as unavailable:
> 
> ```swift
> @available(*, unavailable, message: "This subscript is unavailable in OpenCombine yet")
> public static subscript<EnclosingSelf: AnyObject>(
>     _enclosingInstance object: EnclosingSelf,
>     wrapped wrappedKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Value>,
>     storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Published<Value>>
> ) -> Value {
>     get { fatalError() }
>     set { fatalError() }
> }
> ```

I've added this subscript, but I didn't find that implementation in Apple Combine framework (Xcode 11.0 beta 6). I will be grateful If you give me more information about that functionality and how I can use that. Thanks!

Sources/OpenCombine/Published.swift
broadwaylamb: Please don't mark the conversation as resolved until you commit the changes

broadwaylamb:
@beastgrim sure.

This is used for accessing the object that has its property wrapped in `@Published`. You can find how it can be used [here](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#referencing-the-enclosing-self-in-a-wrapper-type).

You may wonder why I said it's non-trivial. The answer is that you will somehow need to provide a default implementation for the `objectWillChange` property of the `ObservableObject` protocol. According to Apple engineers, [that requires asking the Swift runtime for class metadata](https://forums.swift.org/t/is-there-possible-to-support-inheriting-from-a-generic-type-class-classa-t-classb-t/21541/17).

Sources/OpenCombine/Published.swift
beastgrim: For now I removed this initialisation because that makes framework uncompilable in use.

Sources/OpenCombine/Published.swift
beastgrim: Now little bit shorter

Sources/OpenCombine/Published.swift
beastgrim: This makes framework uncompilable. Are you sure that @available line should be uncommented?

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        fileprivate func send(_ input: Output) {
+            subject.send(input)
+        }
+    }
+
+    /// The property that can be accessed with the
+    /// `$` syntax and allows access to the `Publisher`
+    public let projectedValue: Published<Value>.Publisher
+
```
OpenCombineBot: - :no_entry_sign: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/OpenCombine/Published.swift
broadwaylamb: Yes, pretty sure, I've just successfully compiled your branch with the attribute uncommented. Maybe update your Xcode to beta 7?

Sources/OpenCombine/Published.swift
broadwaylamb: No, I mean, not narrower lines, but less words. You can just copy the code from my previous comment (and preserve the indentation)

Sources/OpenCombine/Published.swift
broadwaylamb: Try this in Xcode beta 7.

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  Published.swift
+//  OpenCombine
+//
+//  Created by Евгений Богомолов on 01/09/2019.
+//
+
+#if swift(>=5.1)
+@propertyWrapper public struct Published<Value> {
```
broadwaylamb, line 15: Please add documentation comment for the struct:

```swift
/// Adds a `Publisher` to a property.
///
/// Properties annotated with `@Published` contain both the stored value
/// and a publisher which sends any new values after the property value
/// has been sent. New subscribers will receive the current value
/// of the property first.
```

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,80 @@
...
+            subject.subscribe(subscriber)
+        }
+
+        fileprivate let subject: OpenCombine.CurrentValueSubject<Value, Never>
+
+        fileprivate init(_ output: Output) {
+            self.subject = .init(output)
+        }
+
+        fileprivate func send(_ input: Output) {
```
broadwaylamb: This method is unnecessary, since `subject` is now fileprivate.

Sources/OpenCombine/Published.swift
```diff
@@ -0,0 +1,80 @@
...
+            subject.send(input)
+        }
+    }
+
+    /// The property that can be accessed with the
+    /// `$` syntax and allows access to the `Publisher`
+    public let projectedValue: Published<Value>.Publisher
```
broadwaylamb: This should be lazy var. I've checked, in Combine it **is** lazy var. This publisher should not be created until somebody reads `projectedValue` for the first time. So we need to make it lazy, like it was before:

```swift
private var _value: Value
public private(set) lazy var projectedValue = Publisher(self._value)
```

You can use the `dump` function from the standard library to explore the behavior of Combine:

```swift
import Combine

var published = Published(initialValue: 42)
dump(published)
// Prints:
// ▿ Combine.Published<Swift.Int>
//   - value: 42
//   - publisher: nil
//   - objectWillChange: nil

_ = published.projectedValue

dump(published)
// Prints:
// ▿ Combine.Published<Swift.Int>
//   - value: 42
//   ▹ publisher: Optional(Combine.Published<Swift.Int>.Publisher(subject: Combine.CurrentValueSubject<Swift.Int, Swift.Never>))
//   - objectWillChange: nil
```

beastgrim: I've checked this too, and it's turned out that projectedValue not used lazy storage, so I created publisher property which projectedValue wraps.

Sources/OpenCombine/Published.swift
beastgrim: That's right now

Sources/OpenCombine/Published.swift
beastgrim: Done

Sources/OpenCombine/Published.swift
```diff
@@ -43,28 +54,40 @@
...
     /// `$` syntax and allows access to the `Publisher`
-    public let projectedValue: Published<Value>.Publisher
-    
-    /* Makes framework uncompilable
+    public var projectedValue: Publisher {
+        mutating get {
+            if publisher == nil {
+                publisher = Publisher(value)
+            }
+            return publisher!
```
broadwaylamb: Force unwrap can be avoided here:

```swift
if let publisher = publisher {
    return publisher
}

let publisher = Publisher(value)
self.publisher = publisher
return publisher
```

Sources/OpenCombine/Published.swift
```diff
@@ -43,28 +54,40 @@
...
         "@Published is only available on properties of classes")
-    */
+
     public var wrappedValue: Value {
-        get { projectedValue.subject.value }
-        set { projectedValue.subject.value = newValue }
+        get { value }
+        set {
+            value = newValue
+            publisher?.send(newValue)
```
broadwaylamb: We can just set:

```swift
publisher?.subject.value = newValue
```

And please remove the `send` method.

Sources/OpenCombine/Published.swift
```diff
@@ -42,11 +42,8 @@
         /// - Parameters:
         ///     - subscriber: The subscriber to attach to this `Publisher`.
         ///                   once attached it can begin to receive values.
-        public func receive<SubscriberType>
-            (subscriber: SubscriberType)
-            where Value == SubscriberType.Input,
-            SubscriberType: Subscriber,
-            SubscriberType.Failure == Published<Value>.Publisher.Failure
+        public func receive<Downstream: Subscriber>(subscriber: Downstream)
+        where Downstream.Input == Value, Downstream.Failure == Never
```
broadwaylamb: Let’s indent this line with 4 more spaces.

Sources/OpenCombine/Published.swift
broadwaylamb: Nice!

broadwaylamb:
Thanks! This looks good to me. Are you up to write some tests? I could write them myself if you don’t feel like it.

We also have our CI failing, but that’s not your fault, it just needs to be updated for the newest Swift version.

beastgrim:
I can write tests for Published but before I would like complete all in Published functionality. There is one property that doesn't exist now "objectWillChange"
```
▿ Combine.Published<Swift.String>
  - value: "test"
  - publisher: nil
  - objectWillChange: nil
```

So I'm going make some investigation how and when this "objectWillChange" uses in Combine framework and after that finish code in "Published" struct. And then cover all code with tests.

I also do not mind if you confirm this pull request. And I will make all these things in another pull request. As it suits you best.

broadwaylamb:
I’ve already investigated it and have a prototype implementation :)

This `objectWillChange` thing is related to the `ObservableObject` protocol and involves some low-level stuff, so we’d need to implement `ObservableObject` first. It doesn’t affect the behavior of `Published` right now anyway, so it will be okay if you write the tests for what’s already done and I’ll merge this PR.

But if you’d really like to try to make it, you can take a look at [this library](https://github.com/wickwirew/Runtime) and [this doc on type metadata](https://github.com/apple/swift/blob/master/docs/ABI/TypeMetadata.rst).

beastgrim:
I could not come up with tests for "Published",  this is more like a template for tests. Sorry 😐

Tests/OpenCombineTests/PublishedTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        ("testPublisher", testPublisher),
+        ("testTestSuiteIncludesAllTests", testTestSuiteIncludesAllTests),
+    ]
+
+    func testPublisher() {
+        #if OPENCOMBINE_COMPATIBILITY_TEST
+        var published: OpenCombine.Published<Int> = Published(initialValue: 0)
+        // All tests for CurrentValueSubject already exists
+        XCTAssert(publisher is CurrentValueSubject)
+        #endif
```
broadwaylamb: This whole method doesn't compile when `OPENCOMBINE_COMPATIBILITY_TEST` is enabled.

You can try it yourself, execute the following command from the root of the project:

```
make generate-compatibility-xcodeproj
```

and then try to run the tests in an iOS simulator.

I think you can just leave this method empty.

Tests/OpenCombineTests/PublishedTests.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // MARK: -
+
+    func testTestSuiteIncludesAllTests() {
```
broadwaylamb: This too

Tests/OpenCombineTests/XCTestManifests.swift
```diff
@@ -42,6 +42,7 @@ public func allTests() -> [XCTestCaseEntry] {
         testCase(SetFailureTypeTests.allTests),
         testCase(SinkTests.allTests),
         testCase(SubscribersDemandTests.allTests),
+        testCase(Publisher.allTests),
```
broadwaylamb: Here you're gonna get merge conflict, you can just delete this whole file.

broadwaylamb (changes requested):
So I've just merged #54, can I ask you to synchronize with the master branch?

Tests/OpenCombineTests/PublishedTests.swift
```diff
@@ -0,0 +1,31 @@
...
+        ("testPublisher", testPublisher),
+        ("testTestSuiteIncludesAllTests", testTestSuiteIncludesAllTests),
+    ]
+
+    func testPublisher() {
+    }
+
+    // MARK: -
+
+    func testTestSuiteIncludesAllTests() {
```
OpenCombineBot: - :no_entry_sign: Empty XCTest method should be avoided. (`empty_xctest_method`)

broadwaylamb (approved):
Thank you for your contribution, @beastgrim!

PR #56, opened 2019-09-12
Fix failing tests caused by Apple changes in GM Seed.

epatey:


broadwaylamb:
Hm, these linter errors are weird, I've explicitly disabled it there:

https://github.com/broadwaylamb/OpenCombine/blob/07c7a98d72cc221e311b8bb0f13e8b62c84d784e/Tests/OpenCombineTests/SubscribersTests/SubscribersDemandTests.swift#L99-L189

epatey:
> Hm, this linter errors are weird, I've explicitly disabled it there:
> 
> https://github.com/broadwaylamb/OpenCombine/blob/07c7a98d72cc221e311b8bb0f13e8b62c84d784e/Tests/OpenCombineTests/SubscribersTests/SubscribersDemandTests.swift#L99-L189

epatey:
Whoops. Didn't mean to close that. Yeah. I deleted the swiftlint disable, because when I run swiftlint locally, it complained that it wasn't needed. Oy. I'll put it back.

broadwaylamb:
@epatey updating SwiftLint should fix that problem for you.

Sources/OpenCombine/AnySubscriber.swift
```diff
@@ -231,24 +231,28 @@ internal final class SubjectSubscriber<Downstream: Subject>
 {
     internal var downstreamSubject: Downstream?
     internal var upstreamSubscription: Subscription?
+    private var isCompleted = false
```
broadwaylamb: I wonder if instead of adding a new flag we can check if `downstreamSubject` is nil, and set it to nil upon completion. I may be wrong here though.

Tests/OpenCombineTests/AnySubscriberTests.swift
```diff
@@ -138,15 +138,22 @@ final class AnySubscriberTests: XCTestCase {
...
-        let expectedShuffledEvents =
-            shuffledEvents.compactMap(subscriberEventToSubjectEvent)
+//        let expectedShuffledEvents =
+//            shuffledEvents.compactMap(subscriberEventToSubjectEvent)
+        // TODO: Because Apple changed the behavior of AnySubscriber to become inactive
+        // after it receives its first `.completion`, this shuffled testing no longer
+        // really makes any sense.
+        // Depending on the original goal of the shuffled testing, a new test could be
+        // written.
+        let expectedShuffledEvents = [TrackingSubject<Int>.Event]()
```
broadwaylamb: The original goal was to make sure that `AnySubscriber` doesn't do anything smart and just passes the events to the underlying subscriber as-is. Since this is no longer the case, we can indeed remove this whole thing.

Tests/OpenCombineTests/AnySubscriberTests.swift
```diff
@@ -230,3 +237,20 @@ private func subscriberEventToSubjectEvent(
         return .completion(c)
     }
 }
+
+@available(macOS 10.15, iOS 13.0, *)
+extension Array {
+    // swiftlint:disable:next generic_type_name
+    func throughFirstCompletion<T>() -> Array where Element == TrackingSubject<T>.Event {
```
broadwaylamb: Probably better replace `T` with `SubjectValue` or `SubjectOutput`, it seems clearer.

Tests/OpenCombineTests/PublisherTests/DropWhileTests.swift
```diff
@@ -225,16 +225,12 @@ final class DropWhileTests: XCTestCase {
...
                                           .completion(.finished)])
 
         publisher.send(completion: .failure(.oops))
         publisher.send(completion: .failure(.oops))
         XCTAssertEqual(tracking.history, [.subscription("DropWhile"),
-                                          .completion(.finished),
-                                          .completion(.finished),
-                                          .completion(.failure(.oops)),
-                                          .completion(.failure(.oops))])
+                                          .completion(.finished)])
```
broadwaylamb, line 233: I like that they've fixed it, makes more sense now.

broadwaylamb (approved):
LGTM, but I have some minor comments.

Tests/OpenCombineTests/PublisherTests/DropWhileTests.swift
epatey: Agreed. In one of my branches, I have a bunch of tests confirming how Apple behaves in all of these "received x after completion" scenarios.

PR #66, opened 2019-09-23
Better Locking internal API

broadwaylamb:


Tests/OpenCombineTests/Helpers/PerformanceTestCase.swift
```diff
@@ -34,6 +34,9 @@ class PerformanceTestCase: GottaGoFast.PerformanceTestCase {
                                    block)
 #endif
     }
+
+    @inline(never)
+    func blackHole<T>(_: T) {}
```
OpenCombineBot: - :no_entry_sign: Generic type name should be between 3 and 20 characters long: 'T' (`generic_type_name`)

PR #67, opened 2019-09-24
Fix lifetime behaviors of PassthroughSubject

epatey:
While trying to reason about some other code by testing it out with a PassthroughSubject, I realized that the subject appeared to not behaving as expected after reaching a terminal state through any means - i.e. cancelled, finished or failed.

The Apple code properly releases references to everything (the publisher, subscription and subscriber) after the subscription becomes terminal.

My fear is that we have this flaw in many of our publisher implementations. I haven't had time to look yet.

While I was in there, I took care of the todo to use a bag/set collection rather than an array.

### DIFF
--- a/Tests/OpenCombineTests/PassthroughSubjectTests.swift
+++ b/Tests/OpenCombineTests/PassthroughSubjectTests.swift
@@ -18,6 +18,43 @@ final class PassthroughSubjectTests: XCTestCase {
 
     private typealias Sut = PassthroughSubject<Int, TestingError>
 
+    func testReleasesEverythingOnTermination() {
+        enum TerminationReason: CaseIterable {
+            case cancelled
+            case finished
+            case failed
+        }
+
+        for reason in TerminationReason.allCases {
+            weak var weakSubscriber: TrackingSubscriber?
+            weak var weakSubject: Sut?
+            weak var weakSubscription: AnyObject?
+
+            do {
+                let subject = Sut()
+                do {
+                    let subscriber = TrackingSubscriber(receiveSubscription: {
+                        weakSubscription = $0 as AnyObject
+                    })
+                    weakSubscriber = subscriber
+                    weakSubject = subject
+
+                    subject.subscribe(subscriber)
+                }
+
+                switch reason {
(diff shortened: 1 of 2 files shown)


Tests/OpenCombineTests/PassthroughSubjectTests.swift
```diff
@@ -18,6 +18,43 @@ final class PassthroughSubjectTests: XCTestCase {
...
+                    subject.subscribe(subscriber)
+                }
+
+                switch reason {
+                case .cancelled: (weakSubscription as? Subscription)?.cancel()
+                case .finished: subject.send(completion: .finished)
+                case .failed: subject.send(completion: .failure(.oops))
+                }
+                if weakSubscriber != nil { XCTFail("Subscriber leaked when reason is \(reason)") }
+                if weakSubscription != nil { XCTFail("Subscription leaked when reason is \(reason)") }
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 102 characters (`line_length`)

broadwaylamb:
Hi! Thank you for pointing this out.

Unfortunately, I’m already working on this stuff for quite some time in [this branch](https://github.com/broadwaylamb/OpenCombine/tree/experiment/subscriber-list), where I’m completely reimplementing PassthoughSubject by reverse engineering Combine. Combine uses this data structure called SubscriberList, and drops to manual reference counting.

I’d be happy to accept your test cases though!

epatey:
Great. I'll remove my fix from the PR leaving just the failing unit test.

I'd love to figure out a way to genericise this sort of test case so that it can easily be used by all of our operator tests. I think we have to figure out a way to consistently and easily test all of the Reactive Streams rules on all of the operators. Currently, it's pretty hit or miss.

MaxDesiatov:
@broadwaylamb sorry for pestering, but have you had any progress on this by any chance? Do you know of any workarounds that could be applied in the meantime? Seems like a pretty nasty issue.

broadwaylamb:
@MaxDesiatov you are right, thanks for reminding me. I’m gonna close this in favor of #147 and then land that.

PR #93, opened 2019-10-25
WIP: ObservableObject

broadwaylamb:
It kind of works, but not in all cases.

Need more tests, especially for interesting configurations like generic classes etc. Maybe `opencombine_enumerate_class_fields` should be tested separately.

After we make sure everything works as expected, it'd be good to remove unneeded C++ code.

PR #97, opened 2019-10-31
Implementation for ObservableObject

broadwaylamb:
(picking up #93)

- [x] Make it work with generic classes.
- [x] Add tests for NSObject subclasses.
- [x] Add tests where a class contains a field of resilient struct type
- [x] Add CI jobs for Apple platforms without Swift runtime (before iOS 12.2/macOS 10.14.4).
- [x] Add CI jobs for Apple platforms that support Swift 5.1.
- [x] Ideally, test on a 32 bit Apple platform.
- [ ] Remove unneeded C++ code.
- [ ] Resolve FIXMEs


Sources/OpenCombine/Published.swift
```diff
@@ -73,28 +68,35 @@
...
+               @Published is only available on properties of classes
+               """)
     public var wrappedValue: Value {
-        get { value }
-        set {
-            value = newValue
-            publisher?.subject.value = newValue
-        }
+        get { fatalError() }
+        set { fatalError() }
```
OpenCombineBot: - :no_entry_sign: Setter value is not used. (`unused_setter_value`)

Tests/OpenCombineTests/ObservableObjectTests.swift
```diff
@@ -0,0 +1,366 @@
...
+    // Foundation.Notification is resilient struct
+    private var note1 = Notification(name: .init("note 1"))
+    @Published var note2 = Notification(name: .init("note 2"))
+
+    init() {
+        print("sizeof Calendar =", MemoryLayout.size(ofValue: note1))
+    }
+}
+
+
```
OpenCombineBot: - :no_entry_sign: Limit vertical whitespace to a single empty line. Currently 2. (`vertical_whitespace`)

Tests/OpenCombineTests/EnumerateClassFieldsTests.swift
```diff
@@ -0,0 +1,335 @@
+//
+//  EnumerateClassFieldsTests.swift
+//  
+//
+//  Created by Sergej Jaskiewicz on 29.11.2019.
+//
+
+import XCTest
+import Foundation
```
OpenCombineBot: - :no_entry_sign: Imports should be sorted. (`sorted_imports`)

secretpromise:
in some way, I had found a simple way to impl observable object, at least in some platform

for swift 5.1 in Published impl we can use generic where clause

```
@propertyWrapper
public struct Published<Value> {
  ...
  public static subscript<EnclosingSelf: ObservableObject>(_enclosingInstance observed: EnclosingSelf, wrapped wrappedKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Value>, storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Published<Value>>) -> Value where EnclosingSelf.ObjectWillChangePublisher == ObservableObjectPublisher {
    get {
      return observed[keyPath: storageKeyPath].value
    }
    set {
      observed.objectWillChange.send()
      observed[keyPath: storageKeyPath].projectedValue.subject.send(newValue)
      observed[keyPath: storageKeyPath].value = newValue
    }
  }
  ...
```

as in apple's platform we use objc to store property in extension ObservableObject
 ```
var observableObjectPublisherKey: String = "ObservableObject.objectWillChange"

extension ObservableObject where Self.ObjectWillChangePublisher == ObservableObjectPublisher {
  /// A publisher that emits before the object has changed.
  public var objectWillChange: ObservableObjectPublisher {
    if let p = objc_getAssociatedObject(self, &observableObjectPublisherKey) as? ObservableObjectPublisher {
      return p
    }
    let publisher = ObservableObjectPublisher()
    objc_setAssociatedObject(self, &observableObjectPublisherKey, publisher, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
    return publisher
  }
}

```

this may not be universal, but it's the simplest way I could figure out

broadwaylamb:
Hi @Wonder-Prime-Singular, thanks for the suggestion.

Unfortunately, none of those would meet our requirements: API compatibility with Combine, and the same efficiency as Combine.

The first one introduces inconsistency with Combine's API. Namely, because of  the `<EnclosingSelf: ObservableObject>` constraint, it is impossible to use `@Published` in a class that does not conform to `ObservableObject`. And it should be possible, because Combine allows it.

The second one doesn't seem very performant. `objc_setAssociatedObject` accesses global state, and for it to be thread-safe it must use synchronization, which in highly contended conditions can be really slow. The solution implemented in this PR doesn't modify any global state, doesn't use any locking, and in fact is quite close to what Combine does.

MaxDesiatov:
Additionaly, I doubt `objc_setAssociatedObject` is available on any non-Apple platform, while supporting those is one of the major goals of OpenCombine in my understanding.

MaxDesiatov:
Ah dang, I'm sorry @broadwaylamb, I was supposed to force push the rebase to my own fork. I don't think anything is lost, but I'd prefer you to force push your own local branch to restore it 🙏

MaxDesiatov:
Actually, no need, sorry for bothering you. GitHub is helpful in higlighting that the previous HEAD was b6aea29, so I'm restoring it to that.

MaxDesiatov:
Would anyone mind if I close this as superseded by #201?

PR #100, opened 2019-11-12
Implement the Record publisher

broadwaylamb:


Sources/OpenCombine/Publishers/Record.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+}
+
+extension Record: Codable where Output: Codable, Failure: Codable {}
+
+extension Record.Recording: Codable where Output: Codable, Failure: Codable {
+
+    private enum CodingKeys: String, CodingKey {
+        case output
```
OpenCombineBot: - :no_entry_sign: Enums should be explicitly assigned their raw values. (`explicit_enum_raw_value`)

PR #107, opened 2019-11-24
Add Future publisher with tests

### LINKED ISSUE #105: Missing `Future` publisher
I've tried to port [GRDBCombine](https://github.com/groue/GRDBCombine) to OpenCombine and it seems like the biggest missing piece there is a lack of the [`Future`](https://developer.apple.com/documentation/combine/future) publisher. Just adding it as a separate issue here for now to hopefully bump its priority on a possible OpenCombine roadmap 🙂

MaxDesiatov:
Resolves #105

### DIFF
--- a/Sources/OpenCombine/Future.swift
+++ b/Sources/OpenCombine/Future.swift
@@ -60,11 +60,11 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
         for subscription in self._subscriptions where !subscription._isCompleted {
             switch result {
             case let .success(output) where subscription._demand > 0:
-               subscription._demand += subscription._downstream?.receive(output) ?? .none
-               subscription._demand -= 1
-               subscription._receive(completion: .finished)
+                subscription._demand += subscription._downstream?.receive(output) ?? .none
+                subscription._demand -= 1
+                subscription._receive(completion: .finished)
             case let .failure(error):
-               subscription._receive(completion: .failure(error))
+                subscription._receive(completion: .failure(error))
 
             // nothing to do if no demand
             default: ()
--- a/Sources/OpenCombine/Future.swift
+++ b/Sources/OpenCombine/Future.swift
@@ -22,6 +22,7 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
     ) {
         attemptToFulfill { result in
             self._lock.do {
+                guard self.result == nil else { return }
                 self.result = result
                 self._publish(result)
             }
--- a/Sources/OpenCombine/Future.swift
+++ b/Sources/OpenCombine/Future.swift
@@ -60,8 +60,8 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
         for subscription in self._subscriptions where !subscription._isCompleted {
             switch result {
             case let .success(output) where subscription._demand > 0:
-                subscription._demand += subscription._downstream?.receive(output) ?? .none
                 subscription._demand -= 1
+                subscription._demand += subscription._downstream?.receive(output) ?? .none
                 subscription._receive(completion: .finished)
             case let .failure(error):
                 subscription._receive(completion: .failure(error))
--- a/Sources/OpenCombine/Future.swift
+++ b/Sources/OpenCombine/Future.swift
@@ -28,6 +28,10 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
         }
     }
 
+    deinit {
+        _lock.deallocate()
+    }
+
     /// This function is called to attach the specified `Subscriber` to this
     /// `Publisher` by `subscribe(_:)`
     ///
--- a/Sources/OpenCombine/Publishers/Future.swift
+++ b/Sources/OpenCombine/Publishers/Future.swift
@@ -18,8 +18,8 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
     private var result: Result<Output, Failure>?
 
     public init(
-        _ attemptToFulfill: @escaping (@escaping Future<Output, Failure>.Promise
-    ) -> Void) {
+        _ attemptToFulfill: @escaping (@escaping Promise) -> Void
+    ) {
         attemptToFulfill { result in
             self._lock.do {
                 self.result = result
--- a/Sources/OpenCombine/Publishers/Future.swift
+++ b/Sources/OpenCombine/Publishers/Future.swift
@@ -17,7 +17,9 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
 
     private var result: Result<Output, Failure>?
 
-    public init(_ attemptToFulfill: @escaping (@escaping Future<Output, Failure>.Promise) -> Void) {
+    public init(
+        _ attemptToFulfill: @escaping (@escaping Future<Output, Failure>.Promise
+    ) -> Void) {
         attemptToFulfill { result in
             self._lock.do {
                 self.result = result
@@ -26,15 +28,16 @@ public final class Future<Output, Failure>: Publisher where Failure: Error {
         }
     }
 
-    /// This function is called to attach the specified `Subscriber` to this `Publisher` by `subscribe(_:)`
+    /// This function is called to attach the specified `Subscriber` to this
+    /// `Publisher` by `subscribe(_:)`
     ///
     /// - SeeAlso: `subscribe(_:)`
     /// - Parameters:
     ///     - subscriber: The subscriber to attach to this `Publisher`.
     ///                   once attached it can begin to receive values.
-    public final func receive<S>(
-        subscriber: S
-    ) where Output == S.Input, Failure == S.Failure, S: Subscriber {
+    public func receive<Downstream: Subscriber>(
+        subscriber: Downstream
+    ) where Output == Downstream.Input, Failure == Downstream.Failure {
         let subscription = Conduit(parent: self,
                                    downstream: AnySubscriber(subscriber))
 
--- a/Sources/OpenCombine/Publishers/Future.swift
+++ b/Sources/OpenCombine/Publishers/Future.swift
@@ -97,7 +97,7 @@ extension Future {
         }
 
         fileprivate func request(_ demand: Subscribers.Demand) {
-            precondition(demand > 0, "demand must not be zero")
+            demand.assertNonZero()
             _parent?._lock.do {
                 _demand += demand
             }
--- a/Sources/OpenCombine/Publishers/Future.swift
+++ b/Sources/OpenCombine/Publishers/Future.swift
@@ -1,0 +1,112 @@
+//
+//  Future.swift
+//  
+//
+//  Created by Max Desiatov on 24/11/2019.
+//
+
+import COpenCombineHelpers
+
+/// A publisher that eventually produces one value and then finishes or fails.
+public final class Future<Output, Failure>: Publisher where Failure: Error {
+
+    public typealias Promise = (Result<Output, Failure>) -> Void
+
+    private let _lock = UnfairRecursiveLock.allocate()
+    private var _subscriptions: [Conduit] = []
+
+    private var result: Result<Output, Failure>?
+
+    public init(_ attemptToFulfill: @escaping (@escaping Future<Output, Failure>.Promise) -> Void) {
+        attemptToFulfill { result in
+            self._lock.do {
+                self.result = result
+                self._publish(result)
+            }
+        }
+    }
+
+    /// This function is called to attach the specified `Subscriber` to this `Publisher` by `subscribe(_:)`
+    ///
+    /// - SeeAlso: `subscribe(_:)`
+    /// - Parameters:
+    ///     - subscriber: The subscriber to attach to this `Publisher`.
+    ///                   once attached it can begin to receive values.
+    public final func receive<S>(
+        subscriber: S
+    ) where Output == S.Input, Failure == S.Failure, S: Subscriber {
+        let subscription = Conduit(parent: self,
+                                   downstream: AnySubscriber(subscriber))
+
+        _subscriptions.append(subscription)
+       subscriber.receive(subscription: subscription)
+    }
+
+    private func _acknowledgeDownstreamDemand() {
+        _lock.do {
+            guard let result = result else { return }
+            _publish(result)
+        }
+    }
+
+    private func _publish(_ result: Result<Output, Failure>) {
+        for subscription in self._subscriptions where !subscription._isCompleted {
+            switch result {
+            case let .success(output) where subscription._demand > 0:
+               subscription._demand += subscription._downstream?.receive(output) ?? .none
+               subscription._demand -= 1
+               subscription._receive(completion: .finished)
+            case let .failure(error):
+               subscription._receive(completion: .failure(error))
+
+            // nothing to do if no demand
+            default: ()
+            }
+        }
+    }
+}
+
+extension Future {
+
+    fileprivate final class Conduit: Subscription {
+
+        fileprivate var _parent: Future<Output, Failure>?
+
+        fileprivate var _downstream: AnySubscriber<Output, Failure>?
+
+        fileprivate var _demand: Subscribers.Demand = .none
+
+        fileprivate var _isCompleted: Bool {
+            return _parent == nil
+        }
+
+        fileprivate init(parent: Future<Output, Failure>,
+                         downstream: AnySubscriber<Output, Failure>) {
+            _parent = parent
+            _downstream = downstream
+        }
+
+        fileprivate func _receive(completion: Subscribers.Completion<Failure>) {
+            if !_isCompleted {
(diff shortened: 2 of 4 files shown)


Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,114 @@
...
+            XCTFail("no value should be returned")
+        })
+
+        wait(for: [expectation], timeout: 5.0)
+        XCTAssertNotNil(cancellable)
+    }
+
+    func testFutureWithinAFlatMap() {
+        let simplePublisher = PassthroughSubject<String, Never>()
+        var outputValue: String? = nil
```
OpenCombineBot: - :no_entry_sign: Initializing an optional variable with nil is redundant. (`redundant_optional_initialization`)

Sources/OpenCombine/Publishers/Future.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    /// This function is called to attach the specified `Subscriber` to this `Publisher` by `subscribe(_:)`
+    ///
+    /// - SeeAlso: `subscribe(_:)`
+    /// - Parameters:
+    ///     - subscriber: The subscriber to attach to this `Publisher`.
+    ///                   once attached it can begin to receive values.
+    public final func receive<S>(
```
OpenCombineBot: - :no_entry_sign: Generic type name should be between 3 and 20 characters long: 'S' (`generic_type_name`)

Sources/OpenCombine/Publishers/Future.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+
+import COpenCombineHelpers
+
+/// A publisher that eventually produces one value and then finishes or fails.
+public final class Future<Output, Failure>: Publisher where Failure: Error {
+
+    public typealias Promise = (Result<Output, Failure>) -> Void
+
+    private let _lock = UnfairRecursiveLock.allocate()
```
broadwaylamb: You never deallocate this lock. It must be deallocated manually. Can you add a `deinit`?

```swift
deinit {
    _lock.deallocate()
}
```

Sources/OpenCombine/Publishers/Future.swift
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+    }
+}
+
+extension Future {
+
+    fileprivate final class Conduit: Subscription {
+
+        fileprivate var _parent: Future<Output, Failure>?
```
broadwaylamb: ~~Looks like we're having a retain cycle here.~~

Sources/OpenCombine/Publishers/Future.swift
```diff
@@ -0,0 +1,115 @@
...
+
+        fileprivate func _receive(completion: Subscribers.Completion<Failure>) {
+            if !_isCompleted {
+                _parent = nil
+                _downstream?.receive(completion: completion)
+            }
+        }
+
+        fileprivate func request(_ demand: Subscribers.Demand) {
+            precondition(demand > 0, "demand must not be zero")
```
broadwaylamb:
```suggestion
            demand.assertNonZero()
```

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,117 @@
...
+        DispatchQueue.global(qos: .background).async {
+            let delay = Int.random(in: 1...3)
+            print(" * making async call (delay of \(delay) seconds)")
+            sleep(UInt32(delay))
+            if sabotage {
+                completionBlock(false, TestFailureCondition.anErrorExample)
+            }
+            completionBlock(true, nil)
+        }
+    }
```
broadwaylamb: I think we should avoid asynchronous testing: it makes tests flaky and slow.

For testing `Future`, the following scenario should suffice:

1. Initialize the future, store the promise in a local variable.
1. After initialization, ensure that the future has not yet received a value and a completion.
1. Resolve the promise using the local variable it's been saved to.
1. Ensure that the future did receive a value and a completion.

This is what I love about Combine: it abstracted away all the asynchrony behind `Scheduler`s, so we can easily test our publishers without turning to `DispatchQueue`s.

broadwaylamb (changes requested):
Huge thanks for taking this!

This looks good, however, I'd like to ask you to move the `Future.swift` file one level up: the `Publishers` directory is for those types that nest in the `Publishers` enum.

Also, this needs a little bit more testing: right now all the tests are asynchronous, and some code paths are not covered, for instance, the `cancel()` method is never called. It would also be good to test the behavior when the promise is resolved multiple times.

Sources/OpenCombine/Publishers/Future.swift
MaxDesiatov: This is a verbatim copy from `PassthroughSubject.swift` by the way, I assume updating it there as well would fix #67

broadwaylamb: Oh. Nevermind, I'm totally wrong. There **should** be a retain cycle here. It's supposed to be broken by calling `cancel()`.

The problem in #67 is that `Coundits` are never removed from the `_subscriptions` array.

MaxDesiatov:
Looks like some CI steps failed on timeout, but I don't have permissions to restart 😕

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        promise?(.failure(TestError()))
+
+        XCTAssertNotNil(error)
+        XCTAssertNotNil(cancellable)
+    }
+
+    func testFutureWithinFlatMap() {
+        var isCompleted = false
+        let simplePublisher = PassthroughSubject<String, Never>()
+        var promise: (() -> ())?
```
OpenCombineBot: - :no_entry_sign: Prefer `-> Void` over `-> ()`. (`void_return`)

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        var promise: (() -> Void)?
+        var outputValue: String?
+
+        let cancellable = simplePublisher
+            .flatMap { name in
+                Future<String, Never> { fulfill in
+                    promise = { fulfill(.success(name)) }
+                }.map { "\($0) foo" }
+            }
+            .sink(receiveCompletion: { err in
```
OpenCombineBot: - :no_entry_sign: Unused parameter "err" in a closure should be replaced with _. (`unused_closure_parameter`)

broadwaylamb:
Can I ask you to use the `TrackingSubscriberBase` class in tests instead of `sink`, and assert on its  `history` property? We don't want to test `sink` along the way.

This class is designed in such a way that it intentionally doesn't do much for you. For example, you need to explicitly request the desired number of elements, while `sink` will always request `.unlimited`, which is not what's always happening, hence we're not testing an important code path.

Also, IMO, this class makes the tests easier to understand.

You can find the example usages of this class pretty much in any test. Here's how we do it for `CurrentValueSubject`:
https://github.com/broadwaylamb/OpenCombine/blob/130125cb66ddb2ab532a5fe3b36af8bbdbbb7a36/Tests/OpenCombineTests/CurrentValueSubjectTests.swift#L230-L265

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,178 @@
...
+            isCompleted = true
+        }, receiveValue: { value in
+            outputValue = value
+        })
+
+        promise?(.success(true))
+
+        XCTAssertTrue(isCompleted)
+        XCTAssertTrue(outputValue)
+        XCTAssertNotNil(cancellable)
```
broadwaylamb: AFAICT `cancellable` is not an optional, so it can never be `nil`.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,178 @@
...
+
+            error = e
+        }, receiveValue: { _ in
+            XCTFail("no value should be returned")
+        })
+
+        promise?(.failure(TestError()))
+
+        XCTAssertNotNil(error)
+        XCTAssertNotNil(cancellable)
```
broadwaylamb: Same here.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,178 @@
...
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+@available(macOS 10.15, iOS 13.0, *)
+final class FutureTests: XCTestCase {
+    struct TestError: Error {}
```
broadwaylamb: We already have the `TestingError` struct for this purpose.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,178 @@
...
+        XCTAssertNil(outputValue)
+        simplePublisher.send("one")
+        promise?()
+
+        XCTAssertEqual(outputValue, "one foo")
+
+        simplePublisher.send(completion: .finished)
+        XCTAssertTrue(isCompleted)
+        XCTAssertNotNil(cancellable)
+    }
```
broadwaylamb: This test feels like it's testing a lot more than just `Future`: it also tests `PassthroughSubject`, `FlatMap`, and `Sink`. I think we should prefer more isolated tests. Here, I'm not sure why we need to test this scenario.

broadwaylamb (changes requested):

Sources/OpenCombine/Future.swift
```diff
@@ -0,0 +1,119 @@
...
+            _publish(result)
+        }
+    }
+
+    private func _publish(_ result: Result<Output, Failure>) {
+        for subscription in self._subscriptions where !subscription._isCompleted {
+            switch result {
+            case let .success(output) where subscription._demand > 0:
+               subscription._demand += subscription._downstream?.receive(output) ?? .none
+               subscription._demand -= 1
```
broadwaylamb: I think these two lines should be swapped. This is how we usually do it, and in the general case it makes sense if calling the `receive` method causes recursion (in which case the decrementing line is only reached after the recursion is terminated, which is bad).

By the way, recursion. This code:

```swift
var promise: Future<Int, Never>.Promise?
let future = Future<Int, Never> {
    promise = $0
}

let cancellable = future.sink {
    promise?(.success($0 + 1))
    print($0)
}

promise?(.success(0))

cancellable.cancel()
```
should not cause infinite recursion (it should just print 0).

After a brief look at your implementation I'm not sure there won't be infinite recursion in this case. Can you please add a test for this behavior? It'd be good to test that there's no infinite recursion not only when receiving value, but also when we're receiving completion.

Sources/OpenCombine/Future.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        fileprivate func _receive(completion: Subscribers.Completion<Failure>) {
+            if !_isCompleted {
+                _parent = nil
+                _downstream?.receive(completion: completion)
+            }
+        }
+
+        fileprivate func request(_ demand: Subscribers.Demand) {
+            demand.assertNonZero()
```
broadwaylamb, line 105: We should test that requesting zero demand causes a crash here. You can use the `assertCrashes` function for this. Search through the project for example usages.

Note that the `assertCrashes` only works with SPM. In Xcode it will never actually report absence of a crash.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,129 @@
...
+
+#if OPENCOMBINE_COMPATIBILITY_TEST
+import Combine
+#else
+import OpenCombine
+#endif
+
+@available(macOS 10.15, iOS 13.0, *)
+final class FutureTests: XCTestCase {
+    private typealias SUT = Future<Int, TestingError>
```
broadwaylamb: Let's rename it to `Sut`  just to have consistent naming across the project.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        var promise: SUT.Promise?
+
+        let future = SUT { promise = $0 }
+
+        let subscriber = TrackingSubscriber(
+            receiveSubscription: { subscription in
+                subscription.request(.unlimited)
+            }, receiveValue: { _ in
+                XCTFail("no value should be returned")
+                return .unlimited
```
broadwaylamb: The conventional thing to do is to return `.none` to indicate that we don't want any more values than we already requested.
```suggestion
                return .none
```

Sources/OpenCombine/Future.swift
MaxDesiatov: Great shout, there is indeed recursion in that case! Swapping the lines didn't help though. By the way, as I mentioned, most of this code is a copy with minor edits from `PassthroughSubject`, which has this:

```swift
let newDemand = subscription._downstream?.receive(input) ?? .none
subscription._demand += newDemand
subscription._demand -= 1
```

broadwaylamb: >Swapping the lines didn't help though.

I never said it should have :)

>By the way, as I mentioned, most of this code is a copy with minor edits from `PassthroughSubject`

The `PassthroughSubject` and `CurrentValueSubject` publishers were one of the first things I implemented in OpenCombine, before I gained any experience with it, so they are in many ways not optimal (and maybe even not totally correct).

I'm working on a better implementation in the [`experiment/subscriber-list`](https://github.com/broadwaylamb/OpenCombine/tree/experiment/subscriber-list) branch. It's not of the highest priority though, and I want #85 to be merged first.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,140 @@
...
+
+        future.subscribe(subscriber)
+
+        XCTAssertEqual(subscriber.history, [
+            .subscription("Future"),
+            .value(42),
+            .completion(.finished)
+        ])
+    }
+
```
OpenCombineBot: - :no_entry_sign: Lines should not have trailing whitespace. (`trailing_whitespace`)

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,140 @@
...
+        future.subscribe(subscriber)
+
+        XCTAssertEqual(subscriber.history, [
+            .subscription("Future"),
+            .value(42),
+            .completion(.finished)
+        ])
+    }
+    
+    func testNoDemand() {
```
broadwaylamb: We already have a number of tests that are called `testCrashesOnZeroDemand`, so let's be consistent.
```suggestion
    func testCrashesOnZeroDemand() {
```

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,158 @@
...
+            guard case .finished = $0 else {
+                XCTFail()
+                return
+            }
+            promise?(.success(42))
+        })
+        future.subscribe(subscriber)
+
+        promise?(.success(0))
+    }
```
broadwaylamb: Let's add `XCTAssertEqual(subscriber.history, [/* whatever */])`

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,158 @@
...
+        }, receiveValue: {
+            promise?(.success($0 + 1))
+            return .none
+        }, receiveCompletion: {
+            guard case .finished = $0 else {
+                XCTFail()
+                return
+            }
+            promise?(.success(42))
+        })
```
broadwaylamb: Let's add one more test for the failure scenario.

Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -0,0 +1,158 @@
...
+        let future = Sut { promise = $0 }
+
+        let subscriber = TrackingSubscriber(receiveSubscription: { subscription in
+            subscription.request(.unlimited)
+        }, receiveValue: {
+            promise?(.success($0 + 1))
+            return .none
+        }, receiveCompletion: {
+            guard case .finished = $0 else {
+                XCTFail()
```
OpenCombineBot: - :no_entry_sign: An XCTFail call should include a description of the assertion. (`xctfail_message`)

broadwaylamb (approved):
Great!

PR #51, opened 2019-08-31
Zip implementation

epatey:

PR #109, opened 2019-11-24
Add Zip implementation

MaxDesiatov:
Looks like #51 is abandoned, this rebases it on top of `master` and cleans up `allTests`.

Closes #51.


MaxDesiatov:
The Linux test failures are quite peculiar, no idea where to start looking though. Is there a recommended way to debug this on macOS? Maybe attaching debugger remotely to a Docker container or something?

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+            // strict given that the documentation says:
+            //      any operation that would result in a negative value is
+            //      clamped to .max(0).
+            //It doesn't say anything about fatalErrors
+            if downstreamDemand != .none {
+                downstreamDemand -= 1
+            }
+            return .sendValueDownstream(value)
+        case .receivedFinishedFromUpstream:
+            return (processingValueForChild != nil || lockedAreMoreValuesPossible()) ? .noAction : .sendFinishDownstream
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 120 characters (`line_length`)

broadwaylamb:
@MaxDesiatov macOS tests are also failing, why not start with those? :)

MaxDesiatov:
What I find interesting is that these tests for `Zip` always pass in a debug build, but fail in release on macOS. I wonder if they would fail in release on iOS as well and it's worth testing both modes (or at least release) on CI?

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,760 @@
...
+            return .sendValueDownstream(value)
+        case .receivedFinishedFromUpstream:
+            return (processingValueForChild != nil || lockedAreMoreValuesPossible()) ?
+                .noAction : .sendFinishDownstream
+        case .receivedRequestFromDownstream(let demand):
+            return .sendRequestUpstream(demand)
+        }
+    }
+
+    @discardableResult private func processQueue() -> Subscribers.Demand? {
```
OpenCombineBot: - :no_entry_sign: Attributes should be on their own lines in functions and types, but on the same line as variables and imports. (`attributes`)

MaxDesiatov:
@broadwaylamb ready for review, I'm open to suggestions as to how to improve the code coverage, especially with those pesky base properties that have `fatalError` in them that can't be tested

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
```diff
@@ -0,0 +1,674 @@
...
+    func testIncreasedDemand() throws {
+        ZipTests.arities.forEach { arity in
+            let (children, zip) = getChildrenAndZipForArity(arity)
+
+            let downstreamSubscriber = TrackingSubscriber(
+                receiveSubscription: {
+                    $0.request(.none)
+                },
+                receiveValue: { _ in
+                    return .max(1)
```
OpenCombineBot: - :no_entry_sign: Prefer implicit returns in closures. (`implicit_return`)

MaxDesiatov:
Hey @broadwaylamb, would you be able to have another look at this PR? All the tests are passing and I'm not sure if this small difference in coverage is acceptable, I don't think that currently uncovered `fatalError` lines are reachable from tests. Except `deinit` which is clearly reachable, can be verified with a debugger, not sure why codecov doesn't track it.

broadwaylamb:
@MaxDesiatov thank you for the PR! I'll take a look as soon as I can. It is not the most trivial one, so I'm afraid it'll take some time for me.

>Except deinit which is clearly reachable, can be verified with a debugger, not sure why codecov doesn't track it.

This is a known issue, yes. I've experienced it myself and spent quite some time trying to understand the object graph before I realized that this is just the coverage issue.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+import COpenCombineHelpers
+#endif
+
+extension Publishers {
+
+    /// A publisher created by applying the zip function to two upstream publishers.
+    public struct Zip<UpstreamA, UpstreamB>: Publisher
+        where UpstreamA: Publisher,
+        UpstreamB: Publisher,
+        UpstreamA.Failure == UpstreamB.Failure
```
broadwaylamb:
```suggestion
    public struct Zip<UpstreamA: Publisher, UpstreamB: Publisher>: Publisher
        where UpstreamA.Failure == UpstreamB.Failure
```

The same goes for all other Zip publishers

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+        ///  `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<Downstream>(subscriber: Downstream)
+            where Downstream: Subscriber,
+            UpstreamB.Failure == Downstream.Failure,
+            Downstream.Input == (UpstreamA.Output, UpstreamB.Output)
```
broadwaylamb:
```suggestion
        public func receive<Downstream: Subscriber>(subscriber: Downstream)
            where Downstream.Failure == Failure, Downstream.Input == Output
```

The same below.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+// note that it does not refer to `Upstream`.
+// This allows `InnerBase` to do most of the heavy lifting without regard to the
+// upstream publisher's value type.
+private protocol ChildSubscription: AnyObject, Subscription {
+    var state: ChildState { get set }
+    var childIndex: Int { get }
+    var hasValue: Bool { get }
+}
+
+fileprivate final class ChildSubscriber<Upstream: Publisher, Downstream: Subscriber>
```
broadwaylamb, line 639: Combine usually uses a struct for such things. See, for example, `SwitchToLatest`:

https://github.com/broadwaylamb/OpenCombine/blob/621f9709980350c3c63db0fa9069c2f391c8701b/Sources/OpenCombine/Publishers/Publishers.SwitchToLatest.swift#L290-L311

That struct holds a *strong* reference to the `Inner`. It is also a `Subscriber`, but not `Subscription`.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+    ///  - All mutable state must only be accessed while `lock` is held.
+    ///  - In order to avoid any deadlock potential, it is absolutely forbidden to have
+    ///      any sort of call out from this class while the lock is held. This is why
+    ///      the draining of the work queue uses a relatively complex pattern.
+    private final let lock = UnfairRecursiveLock.allocate()
+
+    private final let downstream: Downstream
+    private final var downstreamDemand = Subscribers.Demand.none
+    private final var queueIsBeingProcessed = false
+    private final var queuedWork = ArraySlice<QueuedWork>()
```
broadwaylamb: `private` implies `final`.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,756 @@
...
+        case .receivedFinishedFromUpstream:
+            return (processingValueForChild != nil || areMoreValuesPossible) ?
+                .noAction : .sendFinishDownstream
+        case .receivedRequestFromDownstream(let demand):
+            return .sendRequestUpstream(demand)
+        }
+    }
+
+    @discardableResult
+    private func processQueue() -> Subscribers.Demand? {
```
broadwaylamb: This whole thing with the queue was used to avoid recursive locks. However, IMO, this significantly complicates the code. Also, Combine uses recursive locks, so I see no reason not to use them here.

Could you please change your implementation to use recursive locks?

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
```diff
@@ -0,0 +1,692 @@
...
+        }
+
+        return (children, zip)
+    }
+
+    func testImmediateFinishWhenOneChildFinishesWithNoSurplus() {
+        ZipTests.arities.forEach { arity in
+            for childToFinish in (0..<arity) {
+                let description =
+                    String(format:"Zip\(arity) childToFinish=\(childToFinish)")
```
broadwaylamb: Why use this initializer?
```suggestion
                let description = "Zip\(arity) childToFinish=\(childToFinish)"
```

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
```diff
@@ -0,0 +1,692 @@
...
+            receiveSubscription: { $0.request(.unlimited) })
+
+        zip.subscribe(downstreamSubscriber)
+
+        child1Publisher.send(completion: .finished)
+        child2Publisher.send(completion: .finished)
+        child1Publisher.send(completion: .finished)
+
+        XCTAssertEqual(downstreamSubscriber.history, [.subscription("Zip"),
+                                                      .completion(.finished)])    }
```
broadwaylamb:
```suggestion
                                                      .completion(.finished)])
    }
```

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
```diff
@@ -0,0 +1,692 @@
...
+    // subcriptions sends a `.finished`. This means that Apple's behavior is inconsistent.
+    // Sometimes, the `Zip` remains alive even though no futher emissions are possible.
+    // Sometimes it finishes. Ugh.
+    //
+    // If I were in charge, `Zip` would finish as soon as it becomes impossible for it
+    // to send another value - regarless of what triggers that change in state.
+
+    func testZipCompletesOnlyAfterAllChildrenComplete() {
+        let child1Publisher = PassthroughSubject<Int, Never>()
+        let child2Publisher = PassthroughSubject<Int, Never>()
```
broadwaylamb: A lot of tests here use `PassthroughSubject`. It'd be good if you could replace it with `CustomPublisherBase`, so that we could assert on its `CustomSubscription`'s `history`. This way we test more of the observable behavior.

broadwaylamb (changes requested):
Thank you.

Added some comments after a brief look. My main concern is the work queue stuff, which is very hard to understand.

Also, could you please add tests that would test the `CustomStringConvertible`, `CustomReflectable` and `CustomPlaygroundDisplayConvertible` conformances for the `Inner` and `ChildSubscriber` objects? In Combine they usually conform to those protocols, so they should in OpenCombine as well.

For reference you can use `FlatMap` tests:

https://github.com/broadwaylamb/OpenCombine/blob/621f9709980350c3c63db0fa9069c2f391c8701b/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift#L1041-L1082

bioche:
Not sure if this is the right place to put it but I have tested this implementation of Zip and the following code completes without displaying the value : 

```swift
let subject = CurrentValueSubject<Void, Never>(())
C.cancellables.append(
    Publishers.Zip(["hello"].publisher, subject)
              .sink(receiveCompletion: { _ in print("completed") },
                    receiveValue: { print($0) })
)
```

The same code with Apple Combine displays ("hello", ()) but no "completed".
Personally I would have hoped to have both the value and then completed displayed but I'm not sure what is the best options ;)

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,765 @@
...
+        public func receive<Downstream: Subscriber>(subscriber: Downstream) where
+            UpstreamB.Failure == Downstream.Failure,
+            Downstream.Input == (UpstreamA.Output, UpstreamB.Output)
+        {
+            _ = Inner<Downstream>(downstream: subscriber, a, b)
+        }
+    }
+
+    /// A publisher created by applying the zip function to three upstream publishers.
+    public struct Zip3<
```
OpenCombineBot, line 51: - :no_entry_sign: Colons should be next to the identifier of the inheriting type (`inheritance_colon`)

MaxDesiatov: @broadwaylamb there's no way to satisfy both this rule and your requirement for protocol conformance constraints to be within angle brackets instead of `where`. Which one would you choose then?

MaxDesiatov: ...and we also need to fit these within 90 characters to avoid triggering the `line_length` warning.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,765 @@
...
+      UpstreamA: Publisher,
+      UpstreamB: Publisher,
+      UpstreamC: Publisher
+    > : Publisher where
+      UpstreamA.Failure == UpstreamB.Failure,
+      UpstreamB.Failure == UpstreamC.Failure
```
broadwaylamb: How about this?

```suggestion
    public struct Zip3<UpstreamA: Publisher,
                       UpstreamB: Publisher,
                       UpstreamC: Publisher>
        : Publisher
        where UpstreamA.Failure == UpstreamB.Failure,
              UpstreamB.Failure == UpstreamC.Failure
```

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,764 @@
...
+            case .sendValueDownstream(let value):
+                let newDemand = downstream.receive(value)
+                if newDemand != .none {
+                    lock.lock()
+                        downstreamDemand += newDemand
+                        demandReceivedWhileProcessing = newDemand
+                    lock.unlock()
+                }
+            case .sendRequestUpstream(let demand):
+                lock.lock()
```
OpenCombineBot, line 647: - :no_entry_sign: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/OpenCombine/Publishers/Publishers.Zip.swift
MaxDesiatov: I had a closer look and not sure what you mean, the lock used here is already `UnfairRecursiveLock`, is there anything else that you'd like to be changed in this function?

Sources/OpenCombine/Publishers/Publishers.Zip.swift
MaxDesiatov: I had a look at `SwitchToLatest` and I'm a bit confused, shouldn't the `Outer` class there be named `Inner` or maybe some other name? The current name leads to this, note the mismatch between the argument name and the argument type:

```swift
init(inner: Outer, index: UInt64) {
  self.index = index
  self.outer = inner
}
```

I think these names (applicable to `Side` too) don't make their purpose incredibly clear. What exactly is being `Outer` (or `Inner`, or `Side` for that matter) here? A subscription? A subscriber? A publisher?

MaxDesiatov: (I don't mind converting that class to a struct, but it's not as straightforward and I'd like to know your opinion on these issues first so that I can implement it in a better way)

broadwaylamb:
@MaxDesiatov this is weird: @OpenCombineBot has deleted your comments. I haven't done anything.

MaxDesiatov:
Ah, no worries, the main questions I had for you are still there in review comments, I hope you'd have time at some point to have a look at those 🙂

Sources/OpenCombine/Publishers/Publishers.Zip.swift
broadwaylamb: Basically:

https://github.com/OpenCombine/OpenCombine/blob/48d22f455a8edb22d31f5806a98bfeffd1c0cdb8/Sources/OpenCombine/Publishers/Publishers.Zip.swift#L440-L442

The whole queue thing was introduced to avoid calling any client code with the (non-recursive) lock held. But the same can be achieved by just using the recursive lock. Since we now use it, the point of having to deal with the queue is lost. What I suggest is let's get rid of the whole queue thing. It's very hard to understand: just look at the `processQueue()` method.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,698 @@
...
+        return upstreamSubscriptions.allSatisfy { $0.hasValue }
+    }
+
+    private var areMoreValuesPossible: Bool {
+        // More values are possible if all children are (active || have surplus)
+        return upstreamSubscriptions
+            .allSatisfy { $0.state == .active || $0.hasValue }
+    }
+
+
```
OpenCombineBot, line 552: - :no_entry_sign: Limit vertical whitespace to a single empty line. Currently 2. (`vertical_whitespace`)

Sources/OpenCombine/Publishers/Publishers.Zip.swift
MaxDesiatov: Thanks, the queue code has been removed.

Sources/OpenCombine/Publishers/Publishers.Zip.swift
MaxDesiatov: I'm still not sure what would be the purpose of changing that class to a struct and the naming still seems to be inconsistent to me. Can you please elaborate how you'd like it to look and what would be the best naming approach consistent across the codebase?

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
MaxDesiatov: I've changed some of the tests to use `CustomPublisherBase` and `CustomSubscription`, but the rest of them also assert on the history of their downstream subscribers. Adding an upstream subscription has an impact on the downstream subscribers history as well, which I think indicates that the nature of the test changes too? Please let me know how you'd like the rest of those tests to look, or if keeping `PassthroughSubject` in them would be fine otherwise.

Tests/OpenCombineTests/PublisherTests/ZipTests.swift
```diff
@@ -0,0 +1,731 @@
...
+    // subcriptions sends a `.finished`. This means that Apple's behavior is inconsistent.
+    // Sometimes, the `Zip` remains alive even though no futher emissions are possible.
+    // Sometimes it finishes. Ugh.
+    //
+    // If I were in charge, `Zip` would finish as soon as it becomes impossible for it
+    // to send another value - regarless of what triggers that change in state.
+
+    func testZipCompletesOnlyAfterAllChildrenComplete() {
+        let upstreamSubscription = CustomSubscription()
+        let child1Publisher = CustomPublisherBase<Int, Never>(subscription: upstreamSubscription)
```
OpenCombineBot, line 177: - :no_entry_sign: Line should be 90 characters or less: currently 97 characters (`line_length`)

albertopeam:
Hi, 

we are trying to fix the `OpenCombine: execute compatibility tests` but we weren't able to reproduce it. Any tip? How can we run the tests against Combine(not via command line) in xcode?

Thanks

stuaustin:
> Hi,
> 
> we are trying to fix the `OpenCombine: execute compatibility tests` but we weren't able to reproduce it. Any tip? How can we run the tests against Combine(not via command line) in xcode?
> 
> Thanks

If you edit `Project.swift` and change the swiftSettings for "OpenCombineTests", you can run the tests against Combine within Xcode.
You just need to change
```
                .unsafeFlags(["-enable-testing"]),
```
to
```
                .unsafeFlags(["-enable-testing", "-DOPENCOMBINE_COMPATIBILITY_TEST"]),
```

Sources/OpenCombine/Publishers/Publishers.Zip.swift
```diff
@@ -0,0 +1,697 @@
...
+        if upstreamSubscription == nil {
+            upstreamSubscription = subscription
+            parent.receivedSubscription(for: self)
+        } else {
+            subscription.cancel()
+        }
+    }
+
+    fileprivate final func receive(_ input: Input) -> Subscribers.Demand {
+        return parent.receivedChildValue(child: self) { values.append(input) }
```
albertopeam, line 691: Hi, I am having a crash when using Zip.

The reason seems to be the unowned parent, when `receive(_ input: Input)` gets called the parent was deallocated and the app crashes.
```swift
fileprivate final func receive(_ input: Input) -> Subscribers.Demand {
    return parent.receivedChildValue(child: self) { values.append(input) }
}
```
Using weak fixes the problem(because of the circular dependency between Publishers.Zip.Inner and ChildSubscriber)
Snipet with the changes
```swift
fileprivate final class ChildSubscriber<Upstream: Publisher, Downstream: Subscriber>
    where Upstream.Failure == Downstream.Failure
{
    private weak var parent: InnerBase<Downstream>?
}

extension ChildSubscriber: Subscriber {
    fileprivate final func receive(subscription: Subscription) {
        if upstreamSubscription == nil, let parent = parent {
            upstreamSubscription = subscription
            parent.receivedSubscription(for: self)
        } else {
            subscription.cancel()
        }
    }

    fileprivate final func receive(_ input: Input) -> Subscribers.Demand {
        guard let parent = parent else { return .none }
        return parent.receivedChildValue(child: self) { values.append(input) }
    }

    fileprivate final func receive(completion: Subscribers.Completion<Failure>) {
        guard let parent = parent else { return }
        parent.receivedCompletion(completion, forChild: self)
    }
}
```
Could you take a look?
I was trying to build an isolated failing test but I have not been able to reproduce it in a controlled environment.

Thanks in advance :)

MaxDesiatov:
Closing as superseded by https://github.com/OpenCombine/OpenCombine/pull/222

Sources/OpenCombine/Publishers/Publishers.Zip.swift
Kyle-Ye: > I'm still not sure what would be the purpose of changing that class to a struct and the naming still seems to be inconsistent to me. Can you please elaborate how you'd like it to look and what would be the best naming approach consistent across the codebase?

Maybe the best practice is to align with Combine's naming.

PR #111, opened 2019-11-26
Add more tests for the Future publisher

### LINKED ISSUE #105: Missing `Future` publisher

MaxDesiatov:
Fully resolves #105

### DIFF
--- a/Tests/OpenCombineTests/PublisherTests/FutureTests.swift
+++ b/Tests/OpenCombineTests/PublisherTests/FutureTests.swift
@@ -216,13 +216,24 @@ final class FutureTests: XCTestCase {
 
         let future = Sut { promise = $0 }
 
-        let subscriber = TrackingSubscriber()
+        var downstreamSubscription: Subscription?
+        let subscriber = TrackingSubscriber(
+            receiveSubscription: { downstreamSubscription = $0 }
+        )
         future.subscribe(subscriber)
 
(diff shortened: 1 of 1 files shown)


Tests/OpenCombineTests/PublisherTests/FutureTests.swift
```diff
@@ -204,4 +204,25 @@ final class FutureTests: XCTestCase {
...
+
+        let subscriber = TrackingSubscriber()
+        future.subscribe(subscriber)
+
+        promise?(.success(42))
+
+        XCTAssertEqual(subscriber.history, [
+            .subscription("Future")
+        ])
+    }
```
broadwaylamb, line 238: Let's catch the subscription when initializing `TrackingSubscriber`:

```swift
let subscriber = TrackingSubscriber(
    receiveSubscription: { downstreamSubscription = $0 }
)
```

and later make sure that after requesting `.max(1)` the value is sent downstream.

broadwaylamb (approved):
Thank you!

PR #122, opened 2019-12-11
Implement Publishers.ReplaceEmpty

spadafiva:
Hey @broadwaylamb.  I saw that we still needed ReplaceError so i decided to tackle it.  I based a lot of the usage of the `lock` on one of the other files, but it would be good to look that part over specifically since I haven't worked very much with locks in my day to day coding.  Let me know if there's anything missing.

### DIFF
--- a/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
@@ -151,6 +151,12 @@ extension Publishers.ReplaceEmpty {
             demand.assertNonZero()
             lock.lock()
             downstreamRequested = true
+            if finishedWithoutUpstream {
+                lock.unlock()
+                _ = downstream.receive(output)
+                downstream.receive(completion: .finished)
+                return
+            }
             guard case let .subscribed(subscription) = status else {
                 lock.unlock()
                 return
--- a/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
@@ -1,0 +1,131 @@
+//
+//  Publishers.ReplaceEmpty.swift
+//  OpenCombine
+//
+//  Created by Joe Spadafora on 12/10/19.
+//
+
+#if canImport(COpenCombineHelpers)
+import COpenCombineHelpers
+#endif
+
+extension Publisher {
+
+    /// Replaces an empty stream with the provided element.
+    ///
+    /// If the upstream publisher finishes without producing any elements,
+    /// this publisher emits the provided element, then finishes normally.
+    /// - Parameter output: An element to emit when the upstream publisher
+    ///                     finishes without emitting any elements.
+    /// - Returns: A publisher that replaces an empty stream with
+    ///            the provided output element.
+    public func replaceEmpty(with output: Output) -> Publishers.ReplaceEmpty<Self> {
+        return .init(upstream: self, output: output)
+    }
+}
+
+extension Publishers {
+
+    /// A publisher that replaces an empty stream with a provided element.
+    public struct ReplaceEmpty<Upstream>: Publisher where Upstream: Publisher {
+
+        /// The kind of values published by this publisher.
+        public typealias Output = Upstream.Output
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Upstream.Failure
+
+        /// The element to deliver when the upstream publisher finishes without delivering any elements.
+        public let output: Upstream.Output
+
+        /// The publisher from which this publisher receives elements.
+        public let upstream: Upstream
+
+        public init(upstream: Upstream, output: Output) {
+            self.upstream = upstream
+            self.output = output
+        }
+
+        /// This function is called to attach the specified `Subscriber` to this `Publisher` by `subscribe(_:)`
+        ///
+        /// - SeeAlso: `subscribe(_:)`
+        /// - Parameters:
+        ///     - subscriber: The subscriber to attach to this `Publisher`.
+        ///                   once attached it can begin to receive values.
+        public func receive<Downstream: Subscriber>(subscriber: Downstream)
+            where Upstream.Failure == Downstream.Failure,
+                  Upstream.Output == Downstream.Input
+        {
+            let inner = Inner(downstream: subscriber, output: output)
+            upstream.subscribe(inner)
+            subscriber.receive(subscription: inner)
+        }
+    }
+}
+
+extension Publishers.ReplaceEmpty {
+
+    private final class Inner<Downstream: Subscriber>
+        : Subscriber,
+          Subscription,
+          CustomStringConvertible,
+          CustomReflectable,
+          CustomPlaygroundDisplayConvertible
+        where Upstream.Failure == Downstream.Failure,
+              Upstream.Output == Downstream.Input
+    {
+
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        private let output: Upstream.Output
+        private let downstream: Downstream
+        private var subscription: Subscription?
+
+        var isEmpty = true
+
+        fileprivate init(downstream: Downstream, output: Upstream.Output) {
+            self.downstream = downstream
+            self.output = output
+        }
+
+        func receive(subscription: Subscription) {
+            self.subscription = subscription
+        }
+
+        func receive(_ input: Output) -> Subscribers.Demand {
+            isEmpty = false
+            return downstream.receive(input)
+        }
+
+        func receive(completion: Subscribers.Completion<Upstream.Failure>) {
+            switch completion {
+            case .finished:
+                if isEmpty {
+                    _ = downstream.receive(output)
+                }
+                downstream.receive(completion: .finished)
+            case .failure:
+                downstream.receive(completion: completion)
+            }
+        }
+
+        func request(_ demand: Subscribers.Demand) {
+            subscription?.request(demand)
+        }
+
+        func cancel() {
+            subscription?.cancel()
+        }
+
+        var description: String { return "ReplaceEmpty" }
+
+        var customMirror: Mirror {
+            return Mirror(self, children: EmptyCollection())
+        }
+
+        var playgroundDescription: Any { return description }
+    }
+}
--- a/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
+++ b/Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
@@ -62,7 +62,6 @@ extension Publishers {
         {
             let inner = Inner(downstream: subscriber, output: output)
             upstream.subscribe(inner)
-            subscriber.receive(subscription: inner)
         }
     }
 }
@@ -86,11 +85,8 @@ extension Publishers.ReplaceEmpty {
         private let downstream: Downstream
 
         private var status = SubscriptionStatus.awaitingSubscription
-        private var terminated = false
-        private var pendingDemand = Subscribers.Demand.none
         private var lock = UnfairLock.allocate()
         private var isEmpty = true
-        private var requestedDemand = false
 
         fileprivate init(downstream: Downstream, output: Output) {
             self.downstream = downstream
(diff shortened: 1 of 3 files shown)


Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,188 @@
...
+
+        /// The kind of values published by this publisher.
+        public typealias Output = Upstream.Output
+
+        /// The kind of errors this publisher might publish.
+        ///
+        /// Use `Never` if this `Publisher` does not publish errors.
+        public typealias Failure = Upstream.Failure
+
+        /// The element to deliver when the upstream publisher finishes without delivering any elements.
```
OpenCombineBot: - :no_entry_sign: Line should be 90 characters or less: currently 104 characters (`line_length`)

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,177 @@
...
+    ///            the provided output element.
+    public func replaceEmpty(with output: Output) -> Publishers.ReplaceEmpty<Self> {
+        return .init(upstream: self, output: output)
+    }
+}
+
+extension Publishers {
+
+    /// A publisher that replaces an empty stream with a provided element.
+    public struct ReplaceEmpty<Upstream>: Publisher where Upstream: Publisher {
```
broadwaylamb:
```suggestion
    public struct ReplaceEmpty<Upstream: Publisher>: Publisher {
```

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,177 @@
...
+
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        private let output: Output
+        private let downstream: Downstream
+
+        private var status = SubscriptionStatus.awaitingSubscription
+        private var lock = UnfairLock.allocate()
+        private var isEmpty = true
```
broadwaylamb: Using [SuperMirror](https://github.com/broadwaylamb/SuperMirror) I managed to look at all the fields that Combine has here:

- `output: Output`
- `downstream: Downstream`
- `receivedUpstream: Bool` (`isEmpty` in your implementation)
- `lock: UnfairLock`
- `downstreamRequested: Bool`
- `finishedWithoutUpstream: Bool`
- `status: SubscriptionStatus`

1. `downstreamRequested` is used to decide whether to send the extra value downstream. If the upstream has finished without producing a value, **and** downstream hasn't requested anything yet, we must not send `output` downstream, we rather should wait for a request. Could you please add a test for it and tweak your implementation for the test to pass?

2. `finishedWithoutUpstream` seems to be set to `true` when we receive completion from the upstream, but don't have any demand from the downstream yet.

3. Can I ask you to rename the existing stored properties so that they match Combine's names, and also add and leverage those flags that I mentioned? I'd also be good to preserve the order of the stored properties. Why? First of all, we want to stay as close to Combine as possible implementation-wise. Also, the other goal of OpenCombine that I have in mind is that it should be a good reference for the people curious about how Combine works. Third, by speaking in the same "language" as Combine, we can make debugging easier: if something works in Combine and doesn't work in OpenCombine, it'd be good if we could inspect state of Combine objects (the values of their stored properties) and immediately map that state to OpenCombine's to see what went wrong.

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,177 @@
...
+                lock.unlock()
+                return .none
+            }
+            isEmpty = false
+            lock.unlock()
+            let demand = downstream.receive(input)
+            guard demand > 0 else {
+                return .none
+            }
+            return demand
```
broadwaylamb: These lines are equivalent to just `return downstream.receive(input)`.

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,177 @@
...
+                    _ = downstream.receive(output)
+                }
+                downstream.receive(completion: .finished)
+            case .failure:
+                downstream.receive(completion: completion)
+            }
+        }
+
+        func request(_ demand: Subscribers.Demand) {
+            demand.assertNonZero()
```
broadwaylamb, line 156: We should test that we actually crash when requesting zero demand.

You can use the `assertCrashes` function for it. See [an example](https://github.com/broadwaylamb/OpenCombine/blob/899a04bb3fc88a855196a0ce90d52b6f1cf5e04c/Tests/OpenCombineTests/PublisherTests/FlatMapTests.swift#L212-L223).

Tests/OpenCombineTests/PublisherTests/ReplaceEmptyTests.swift
```diff
@@ -0,0 +1,277 @@
...
+            .receive(subscription: helper.subscription)
+
+        XCTAssertEqual(helper.subscription.history, [.requested(.unlimited), .cancelled])
+
+        try XCTUnwrap(helper.downstreamSubscription).cancel()
+
+        XCTAssertEqual(helper.subscription.history, [.requested(.unlimited),
+                                                     .cancelled,
+                                                     .cancelled])
+    }
```
broadwaylamb, line 284: Could you leverage the existing function for testing this declared in [CommontTests.swift](https://github.com/broadwaylamb/OpenCombine/blob/899a04bb3fc88a855196a0ce90d52b6f1cf5e04c/Tests/OpenCombineTests/Helpers/CommonTests.swift#L140-L167)?

Also, I'd like to ask you to add the following tests from that file:

- `testReceiveValueBeforeSubscription`
- `testReceiveCompletionBeforeSubscription`
- `testRequestBeforeSubscription`
- `testCancelBeforeSubscription`

For example usage you can refer to [this](https://github.com/broadwaylamb/OpenCombine/blob/899a04bb3fc88a855196a0ce90d52b6f1cf5e04c/Tests/OpenCombineTests/PublisherTests/DelayTests.swift#L417-L447).

Also, there's one more test missing: `testLifecycle` from [TestLifecycle.swift](https://github.com/broadwaylamb/OpenCombine/blob/899a04bb3fc88a855196a0ce90d52b6f1cf5e04c/Tests/OpenCombineTests/Helpers/TestLifecycle.swift#L17-L153). Example [here](https://github.com/broadwaylamb/OpenCombine/blob/899a04bb3fc88a855196a0ce90d52b6f1cf5e04c/Tests/OpenCombineTests/PublisherTests/HandleEventsTests.swift#L232-L236).

broadwaylamb (changes requested):
Thank you! This looks very good.

There are some corner cases to consider though.

Tests/OpenCombineTests/PublisherTests/ReplaceEmptyTests.swift
spadafiva: @broadwaylamb In this case, there was a different behavior for the receives subscription twice than the usual: 

```
CommonTests.swift:150: error: -[OpenCombineTests.ReplaceEmptyTests testReceiveSubscriptionTwice] : XCTAssertEqual failed: ("[.requested(.unlimited)]") is not equal to ("[]")
CommonTests.swift:162: error: -[OpenCombineTests.ReplaceEmptyTests testReceiveSubscriptionTwice] : XCTAssertEqual failed: ("[.requested(.unlimited), .cancelled]") is not equal to ("[.cancelled]")
CommonTests.swift:166: error: -[OpenCombineTests.ReplaceEmptyTests testReceiveSubscriptionTwice] : XCTAssertEqual failed: ("[.requested(.unlimited), .cancelled, .cancelled]") is not equal to ("[.cancelled, .cancelled]")
```

I will add the other following tests from the file though.

broadwaylamb: Are you experiencing this behavior with Combine too?

broadwaylamb: Oh, I see. Okay then, let's leave this test as-is.

spadafiva: I am experiencing it when running the tests with the compatibility flag.

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,181 @@
...
+            guard case .subscribed = status else {
+                lock.unlock()
+                return
+            }
+            status = .terminal
+            lock.unlock()
+            switch completion {
+            case .finished:
+                guard downstreamRequested else {
+                    finishedWithoutUpstream = true
```
broadwaylamb: 👍

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
spadafiva: @broadwaylamb I'm think I've got 1 and 3 covered, but I'm not sure if there is a valid test that we could use for 2.  Let me know if you have any suggestions, but otherwise I've renamed and added everything.

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,181 @@
...
+                downstream.receive(completion: .finished)
+            case .failure:
+                downstream.receive(completion: completion)
+            }
+        }
+
+        func request(_ demand: Subscribers.Demand) {
+            demand.assertNonZero()
+            lock.lock()
+            downstreamRequested = true
```
broadwaylamb, line 158: We should check if `finishedWithoutUpstream` is `true` here, and if it is, send `output` and completion downstream.

```swift
if finishedWithoutUpstream {
    lock.unlock()
    _ = downstream.receive(output)
    downstream.receive(completion: .finished)
    return
}
```

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,181 @@
...
+              Upstream.Output == Downstream.Input
+    {
+
+        typealias Input = Upstream.Output
+        typealias Failure = Upstream.Failure
+
+        private let output: Output
+        private let downstream: Downstream
+
+        private var receivedUpstream = true
```
broadwaylamb: It probably makes sense to invert the value. It used to be `isEmpty`, and now its meaning is the opposite.

Sources/OpenCombine/Publishers/Publishers.ReplaceEmpty.swift
```diff
@@ -0,0 +1,181 @@
...
+            lock.unlock()
+            switch completion {
+            case .finished:
+                guard downstreamRequested else {
+                    finishedWithoutUpstream = true
+                    return
+                }
+                if receivedUpstream {
+                    _ = downstream.receive(output)
+                }
```
broadwaylamb: Oops. Reading from and writing to `downstreamRequested`, `finishedWithoutUpstream` and `receivedUpstream` should be locked with `lock`.

Here's what I suggest:

```swift
func receive(completion: Subscribers.Completion<Upstream.Failure>) {
    lock.lock()
    guard case .subscribed = status else {
        lock.unlock()
        return
    }
    status = .terminal
    if receivedUpstream {
        lock.unlock()
        downstream.receive(completion: completion)
        return
    }
    switch completion {
    case .finished:
        lock.unlock()
        if downstreamRequested {
            lock.unlock()
            _ = downstream.receive(output)
            downstream.receive(completion: completion)
            return
        }
        finishedWithoutUpstream = true
        lock.unlock()
    case .failure:
        lock.unlock()
        downstream.receive(completion: completion)
    }
}
```

broadwaylamb (changes requested):

Tests/OpenCombineTests/PublisherTests/ReplaceEmptyTests.swift
```diff
@@ -0,0 +1,341 @@
...
+
+    func testReplaceEmptyCancelBeforeSubscription() {
+        testCancelBeforeSubscription(inputType: Int.self, shouldCrash: false) {
+            $0.replaceEmpty(with: 1337)
+        }
+    }
+
+    func testReplaceEmptyLifecycle() throws {
+        // TODO: Fix this test to match combine behavior
```
broadwaylamb: What is the issue with this?

spadafiva: That was something I added a few commits ago because I was pushing a test that was failing, but fixed it in a more recent commit

broadwaylamb (approved):
Awesome, thanks!

PR #123, opened 2019-12-11
Update for Xcode 11.3

broadwaylamb:
- Send subscription synchronously in `ReceiveOn` and `Delay` operators
- Some locks made recursive, as they should be
- `ObservableObjectPublisher` doesn't use `PassthroughSubject` under the hood anymore


broadwaylamb:
Let's wait until CircleCI supports Xcode 11.3, shouldn't be too long.

PR #127, opened 2019-12-13
Implement URLSession.DataTaskPublisher

### LINKED ISSUE #124: URLSession.dataTaskPublisher for OpenCombine?
Hi there

So I am on the home stretch, but I have a few issues. Firstly:
I am getting 
```
error: value of type 'URLSession' has no member 'dataTaskPublisher'
            self.urlSession.dataTaskPublisher(for: request)
```

Is there an OpenCombine equivalent to use `dataTaskPublisher`?

Secondly, I am also getting:
```
error: value of type 'Any' has no member 'tryMap'
                .tryMap { (data, response) -> Data in
```
Is this also functional in OpenCombine?

Thanks once again 🙏🏻

broadwaylamb:
Fixes #124

### DIFF
--- a/Sources/OpenCombineFoundation/Helpers/Violations.swift
+++ b/Sources/OpenCombineFoundation/Helpers/Violations.swift
@@ -1,0 +1,17 @@
+//
+//  Violations.swift
+//  
+//
+//  Created by Sergej Jaskiewicz on 13.12.2019.
+//
+
+import OpenCombine
+
+extension Subscribers.Demand {
+    internal func assertNonZero(file: StaticString = #file,
+                                line: UInt = #line) {
+        if self == .none {
+            fatalError("API Violation: demand must not be zero", file: file, line: line)
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


Sources/OpenCombineFoundation/Helpers/Violations.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Sergej Jaskiewicz on 13.12.2019.
+//
+
+import OpenCombine
+
+extension Subscribers.Demand {
+    internal func assertNonZero(file: StaticString = #file,
+                                line: UInt = #line) {
+        if self == .none {
```
MaxDesiatov, line 13:
```suggestion
        guard self != .none else {
```

guarantees that negative case is actually handled this way

broadwaylamb: I have this weird preference to not use `guard` with negation. In my opinion, it makes code harder to read.

PR #128, opened 2019-12-13
Expose OpenCombineFoundation target as a product

MaxDesiatov:
This should fix `import OpenCombineFoundation` issue reported in #124.


broadwaylamb (approved):
How stupid of me.

PR #145, opened 2020-01-23
work around SR-11680

mayoff:
The Swift bug report: https://bugs.swift.org/browse/SR-11680

Swift nightly toolchains are available here: https://swift.org/download/

The Swift nightly toolchains cannot build OpenCombine. Here's why:

1.  The COpenCombineHelpers target defines a non-static function
    (opencombine_stop_in_debugger) in a header file. This function is
    emitted in the target's IR, but not in the target's TBD.

2.  Swift nightly toolchains have assertions enabled, so they use the
    `-validate-tbd-against-ir=missing` build setting. This build setting
    makes the compiler fail if the TBD doesn't match the IR.

This commit turns off the build setting for the OpenCombine target.

### DIFF
--- a/Sources/COpenCombineHelpers/COpenCombineHelpers.cpp
+++ b/Sources/COpenCombineHelpers/COpenCombineHelpers.cpp
@@ -11,6 +11,7 @@
 #include <cstdlib>
 #include <system_error>
 #include <pthread.h>
+#include <signal.h>
 
 #ifdef __APPLE__
 #include <os/lock.h>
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
The linked SR-11680 has different environment specified, namely Xcode 11.0, swift-5.1-RELEASE. Does that mean that using nightlies is not required to reproduce it?

broadwaylamb:
Hi @mayoff, thanks for pointing out this issue!

However, we can't use the `unsafeFlags` in a public target because of [this](https://developer.apple.com/documentation/swift_packages/swiftsetting/3112770-unsafeflags):

>As some build flags can be exploited for unsupported or malicious behavior, a product can’t be used as a dependency in another package if one of its targets uses unsafe flags.

So, basically, clients won't be able to depend on OpenCombine.

What I would suggest to do is simply define `opencombine_stop_in_debugger` out of line in `COpenCombineHelpers.cpp` instead of the header file.

mayoff:
@MaxDesiatov I don't know if the RELEASE builds at https://swift.org/download/ are built with assertions. You could (I assume) trigger the same error by manually setting`-validate-tbd-against-ir=missing` with any toolchain.

@broadwaylamb Oh, good point. I'll revise the patch to non-inline the function.

broadwaylamb (approved):
Nice!

mayoff:
I have rewritten the patch so that it no longer uses `.unsafeFlags`.

PR #146, opened 2020-01-24
add an lldb type summary for Demand

mayoff:
This commit adds a type summary for lldb.

- In the `lldb` console, it formats a `Demand` as `max(N)` or `unlimited`.
- In the Xcode debugger's Variables View, it shows `max(N)` or `unlimited` as the value of the `Demand`, and you don't have to click the disclosure triangle by the `Demand` variable to see the vaule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,3 +89,16 @@ GYB template files have the `.gyb` extension. Run `make gyb` to generate Swift c
 templates. The generated files are prefixed with `GENERATED-`  and are checked into source control. Those
 files should never be edited directly. Instead, the `.gyb` template should be edited, and after that the files
 should be regenerated using `make gyb`.
+
+#### Debugger Support
+
+The file `opencombine_lldb.py`  defines some `lldb` type summaries for easier debugging. These type summaries improve the way `lldb` and Xcode display some OpenCombine values.
+
+To use `opencombine_lldb.py`, figure out its full path. Let's say the full path is `~/projects/OpenCombine/opencombine_lldb.py`. Then the following statement to your `~/.lldbinit` file:
+
+    command script import ~/projects/OpenCombine/opencombine_lldb.py
+
+Currently, `opencombine_lldb.py` defines type summaries for these types:
+
+- `Subscribers.Demand`
+- That's all for now.
--- a/opencombine_lldb.py
+++ b/opencombine_lldb.py
@@ -1,0 +1,29 @@
+# To use `opencombine_lldb.py`, figure out its full path.
+# Let's say the full path is `~/projects/OpenCombine/opencombine_lldb.py`.
+# Then add the following statement to your `~/.lldbinit` file:
+#
+#     command script import ~/projects/OpenCombine/opencombine_lldb.py
+
+import lldb
+
+# Show a Demand as either `max(N)` or `unlimited`.
+def summary_Demand(sb_value, internal_dict):
+    child = sb_value.GetChildAtIndex(0)
+    if not child.IsValid():
+        return 'failed to get child of Demand'
+    number = child.GetValueAsUnsigned()
+
+    # .unlimited is represented by a rawValue of UInt(Int.max) + 1.
+    # Int.max is either 2**31 - 1 or 2**63 - 1 depending on the
+    # target platform. So .unlimited is either 2**31 or 2**63.
+    #     31 = 4 * 8 - 1
+    #     63 = 8 * 8 - 1
+    unlimited = 2**(child.GetByteSize() * 8 - 1)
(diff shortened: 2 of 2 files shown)


OpenCombine.py
```diff
@@ -0,0 +1,21 @@
...
+
+import lldb
+
+# Show a Demand as either `max(N)` or `unlimited`.
+def summary_Demand(sb_value, internal_dict):
+    child = sb_value.GetChildAtIndex(0)
+    if not child.IsValid():
+        return 'failed to get child of Demand'
+    number = child.GetValueAsUnsigned()
+    if number == 9223372036854775808:
```
broadwaylamb: This number would be different on 32 bit platforms, is it possible to handle that?

broadwaylamb (approved):
Cool!

OpenCombine.py
mayoff: Good catch. I'll have to do a little digging to figure out the LLDB API that reports the target's word size.

mayoff:
I've updated the patch to handle 32-bit correctly, I believe. I've only tested on 64-bit though.

OpenCombine.py
```diff
@@ -0,0 +1,22 @@
...
+import lldb
+
+# Show a Demand as either `max(N)` or `unlimited`.
+def summary_Demand(sb_value, internal_dict):
+    child = sb_value.GetChildAtIndex(0)
+    if not child.IsValid():
+        return 'failed to get child of Demand'
+    number = child.GetValueAsUnsigned()
+    size = child.GetByteSize()
+    if number == 2**(8 * size - 1):
```
broadwaylamb: Let's make it clear what's going on:
```suggestion
    if number == 2**(8 * size - 1): # UInt(Int.max) + 1
```

OpenCombine.py
```diff
@@ -0,0 +1,22 @@
+# To use `OpenCombine.py`, figure out its full path.
+# Let's say the full path is `~/projects/OpenCombine/OpenCombine.py`.
+# Then add the following statement to your `~/.lldbinit` file:
+#
+#     command script import ~/projects/OpenCombine/OpenCombine.py
```
broadwaylamb: Can I ask to rename this file according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names)?

Something like `opencombine_lldb.py` so it's immediately clear that this has something to do with LLDB.

P. S. Don't forget to update README too :)

broadwaylamb (changes requested):

OpenCombine.py
```diff
@@ -0,0 +1,22 @@
...
+
+import lldb
+
+# Show a Demand as either `max(N)` or `unlimited`.
+def summary_Demand(sb_value, internal_dict):
```
broadwaylamb:
```suggestion
def summary_demand(sb_value, internal_dict):
```

[PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

OpenCombine.py
```diff
@@ -0,0 +1,22 @@
...
+        return 'failed to get child of Demand'
+    number = child.GetValueAsUnsigned()
+    size = child.GetByteSize()
+    if number == 2**(8 * size - 1):
+        return 'unlimited'
+    else:
+        return 'max(%d)' % number
+
+def __lldb_init_module(debugger, internal_dict):
+    debugger.HandleCommand('type summary add -w swift OpenCombine.Subscribers.Demand -F "' + __name__ + '.summary_Demand"')
```
broadwaylamb:
```suggestion
    debugger.HandleCommand('type summary add -w swift OpenCombine.Subscribers.Demand -F "' + __name__ + '.summary_demand"')
```

OpenCombine.py
mayoff: I prefer keeping the D capitalized here as this method is specifically referring to the Demand type, which is spelled with a capital D.

OpenCombine.py
mayoff: Same as above.

OpenCombine.py
broadwaylamb: Understood. Okay, it's not like it's crucial here, let's leave it as is then.

OpenCombine.py
mayoff: I've clarified the code in the new patch.

OpenCombine.py
mayoff: I've renamed the file as requested. I updated the comments in the file itself and in `README.md`.

broadwaylamb (approved):

PR #156, opened 2020-05-22
Add TimerPublisher and Timer.publish

MaxDesiatov:
This is copied from [upstream](https://github.com/apple/swift/blob/master/stdlib/public/Darwin/Foundation/Publishers%2BTimer.swift) (preserving the license header) with minor changes to resolve namespacing issues and to use our `__UnfairLock`.

@broadwaylamb do you think this would need extensive (if any) test coverage? Since it's almost a direct copy of the Apple's implementation, I would assume the compatibility with upstream is a top priority. Unless you'd like 32-bit compatibility for this, but that could come as a separate PR.


Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,339 @@
...
+        init(parent: TimerPublisher) {
+          lock = Lock.allocate()
+          inner = Inner(parent, self)
+        }
+
+        deinit {
+          lock.deallocate()
+        }
+
+        func addSubsriber<S: Subscriber>(_ sub: S)
```
OpenCombineBot, line 143: - :no_entry_sign: Generic type name should be between 3 and 20 characters long: 'S' (`generic_type_name`)

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,344 @@
...
+
+      private typealias Parent = TimerPublisher
+      private final class Inner<Downstream: Subscriber>: NSObject, Subscription,
+        CustomReflectable, CustomPlaygroundDisplayConvertible
+        where
+        Downstream.Input == Date,
+        Downstream.Failure == Never {
+        private lazy var timer: Timer? = {
+          let t = Timer(timeInterval: parent?.interval ?? 0, repeats: true) {
+            [weak self] _ in self?.timerFired()
```
OpenCombineBot, line 232: - :no_entry_sign: Closure parameters should be on the same line as opening brace. (`closure_parameter_position`)

MaxDesiatov:
@broadwaylamb CI passes now, ready for review 🙂

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+import Foundation
+import OpenCombine
+
+#if canImport(COpenCombineHelpers)
+import COpenCombineHelpers
+#endif
+
+private typealias Lock = __UnfairLock
+
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```
broadwaylamb: I think we should stick with the minimum deployment target that we currently support. I understrand that some APIs used in this implementation are iOS 13-only, but I believe we can work this around by using more low-level CoreFoundation APIs instead. You may refer to this: https://github.com/broadwaylamb/OpenCombine/blob/d680f09932fe68942c3d391d6df296d38b9dcd05/Sources/OpenCombineFoundation/RunLoop%2BScheduler.swift#L168-L205

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// Only support 64bit
+#if !(os(iOS) && (arch(i386) || arch(arm)))
```
broadwaylamb: What prevents us from removing this right now?

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+        self.tolerance = tolerance
+        self.runLoop = runLoop
+        self.mode = mode
+        self.options = options
+      }
+
+      /// Adapter subscription to allow `Timer` to multiplex to multiple subscribers
+      /// the values produced by a single `TimerPublisher.Inner`
+      private class RoutingSubscription: Subscription, Subscriber,
+      CustomStringConvertible, CustomReflectable, CustomPlaygroundDisplayConvertible {
```
broadwaylamb:
```suggestion
      private class RoutingSubscription
          : Subscription,
            Subscriber,
            CustomStringConvertible,
            CustomReflectable,
            CustomPlaygroundDisplayConvertible
      {
```

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+#if !(os(iOS) && (arch(i386) || arch(arm)))
+
+import Foundation
+import OpenCombine
+
+#if canImport(COpenCombineHelpers)
+import COpenCombineHelpers
+#endif
+
+private typealias Lock = __UnfairLock
```
broadwaylamb: We already have the needed typealias here:
https://github.com/broadwaylamb/OpenCombine/blob/d680f09932fe68942c3d391d6df296d38b9dcd05/Sources/OpenCombineFoundation/Helpers/Locking.swift#L14-L15

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+        private var _lockedIsConnected = false
+        var isConnected: Bool {
+          get {
+            lock.lock()
+            defer { lock.unlock() }
+            return _lockedIsConnected
+          }
+
+          set {
+            lock.lock()
```
broadwaylamb:
```suggestion
            return _lockedIsConnected
          }
          set {
            lock.lock()
```

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+          }
+        }
+
+        var description: String { return "Timer" }
+        var customMirror: Mirror { return inner.customMirror }
+        var playgroundDescription: Any { return description }
+        var combineIdentifier: CombineIdentifier { return inner.combineIdentifier }
+
+        init(parent: TimerPublisher) {
+          lock = Lock.allocate()
```
broadwaylamb: Let's move this initialization to the property declaration like we do it in other implementations.

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        deinit {
+          lock.deallocate()
+        }
+
+        func addSubscriber<Sub: Subscriber>(_ sub: Sub)
+          where
+          Sub.Failure == Failure,
+          Sub.Input == Output {
```
broadwaylamb:
```suggestion
        func addSubscriber<Downstream: Subscriber>(_ sub: Downstream)
            where Downstream.Failure == Failure, Downstream.Input == Output
        {
```

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+
+        override var description: String { return "Timer" }
+        var customMirror: Mirror {
+          lock.lock()
+          defer { lock.unlock() }
+          return Mirror(self, children: [
+            "downstream": downstream as Any,
+            "interval": parent?.interval as Any,
+            "tolerance": parent?.tolerance as Any,
+          ])
```
broadwaylamb: Just to be consistent with the rest of our code, where we use arrays instead of dictionaries:
```suggestion
            let children: [Mirror.Child] = [
                ("downstream", downstream as Any),
                ("interval", parent?.interval as Any),
                ("tolerance", parent?.tolerance as Any),
            ]
            return Mirror(self, children: children)
```

Sources/OpenCombineFoundation/Publishers+Timer.swift
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        var playgroundDescription: Any { return description }
+
+        init(_ parent: Parent, _ downstream: Downstream) {
+          lock = Lock.allocate()
```
broadwaylamb: Again, let's move this to the property declaration.

broadwaylamb (changes requested):
Thank you for your work!

I'd like to ask you to indent the code with 4 spaces instead of 2.

The other thing that I think is worth fixing is compatibility with older versions of Apple OSes.

Otherwise it looks good.

broadwaylamb:
Also, yes, if you don't mind, it would be good to add some tests. Tests for timers can be unreliable, I know, but we already have something that is more or less stable:
https://github.com/broadwaylamb/OpenCombine/blob/d680f09932fe68942c3d391d6df296d38b9dcd05/Tests/OpenCombineTests/FoundationTests/RunLoopSchedulerTests.swift#L423-L490

I did't leave a lot of comments there, so feel free to ask questions if you have them.

Sources/OpenCombineFoundation/Publishers+Timer.swift
MaxDesiatov: IDK, would probably need to write some tests and run them on 32-bit platforms to find out 😄

broadwaylamb: BTW, we do run our tests on a 32 bit iOS simulator: https://app.circleci.com/pipelines/github/broadwaylamb/OpenCombine/373/workflows/e6c8e452-f56c-442d-8cc2-b7fceecdbe3c/jobs/1967/steps

MaxDesiatov: Yes, I'm aware of that, thanks 👍

broadwaylamb:
@MaxDesiatov do you mind if I take this over?

MaxDesiatov:
Yes please, sorry for the delay, I've been busy with trying to port OpenCombine to WebAssembly (which mostly works, but I primarily need `ObservableObject`, and your implementation fails to compile, there's a complete rabbit hole with `clang` headers). Also, I put a bit less effort into it than I could, as I was slightly worried Apple could actually sherlock it by open-sourcing Combine at WWDC. I'll look into #109 this week, but feel free to take over this one.

broadwaylamb (approved):

Issue #157, opened 2020-06-02
Result.Inner crash with multiple threads


glessard:
This might be an example of holding it wrong, but it's a multi-threading crash.
```
    func testResultPublisherMultiThreaded() throws
    {
      class ExternallyTriggeredSubscriber: Subscriber
      {
        typealias Input = Int
        typealias Failure = Never

        let combineIdentifier = CombineIdentifier()
        var sub: Subscription?

        func receive(subscription: Subscription) { sub = subscription }
        func receive(_ input: Input) -> Subscribers.Demand { return .none }
        func receive(completion: Subscribers.Completion<Failure>) {}

        func request() { sub?.request(.unlimited) }
        func cancel() { sub?.cancel() }
      }

      for i in 1...1000
      {
        let pub = Result<Int, Never>.OCombine.Publisher(i)
        let sub = ExternallyTriggeredSubscriber()
        pub.subscribe(sub)
        DispatchQueue.global(qos: .utility).async { sub.request() }
        DispatchQueue.global(qos: .utility).async { sub.cancel() }
      }
    }
```

The `downstream` variable should be protected on read and write.

broadwaylamb:
Could you please describe what’s happening exactly? Where does the crash occur?

glessard:
Sorry I don't know what happened earlier, it seemingly auto-posted before I typed anything!

glessard:
To be fair, Apple's version crashes in seemingly the same way. Both fail the contract in that `cancel()` isn't thread-safe.

glessard:
FWIW, this fixes it:
```
---
 .../OpenCombine/Publishers/Result.Publisher.swift  | 14 +++++++++++++-
 1 file changed, 13 insertions(+), 1 deletion(-)

diff --git a/Sources/OpenCombine/Publishers/Result.Publisher.swift b/Sources/OpenCombine/Publishers/Result.Publisher.swift
index 9898ff5..e49febd 100644
--- a/Sources/OpenCombine/Publishers/Result.Publisher.swift
+++ b/Sources/OpenCombine/Publishers/Result.Publisher.swift
@@ -115,6 +115,8 @@ extension Result {
 #endif
 }
 
+import COpenCombineHelpers
+
 extension Result.OCombine {
     private final class Inner<Downstream: Subscriber>
         : Subscription,
@@ -126,6 +128,7 @@ extension Result.OCombine {
         // NOTE: this class has been audited for thread safety.
[...]

PR #158, opened 2020-06-05
Document an expected crash

glessard:
Follow-up to https://github.com/broadwaylamb/OpenCombine/issues/157.

Adds the test with an `assertCrashes` block, to monitor if and when the official Combine stops crashing in this situation (FB7722681).

Note that `Just` and `Optional.Publisher` exhibit the same behaviour.

### DIFF
--- a/Tests/OpenCombineTests/PublisherTests/ResultPublisherTests.swift
+++ b/Tests/OpenCombineTests/PublisherTests/ResultPublisherTests.swift
@@ -144,6 +144,33 @@ final class ResultPublisherTests: XCTestCase {
         """)
     }
 
+    func testResultPublisherMultiThreaded() throws
+    {
+      class ExternallyTriggered: Subscriber {
+        typealias Input = Int
+        typealias Failure = Never
+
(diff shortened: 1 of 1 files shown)


glessard:
The proposed new test trips TSAN on Linux with:
`ERROR: ThreadSanitizer: SEGV on unknown address 0x000000000000`
This is not useful information, as we are expecting it to crash.

broadwaylamb:
Honestly, I don't think this is the right direction. This is a data race, which is undefined behavior, meaning that anything could happen and, strictly speaking, a crash is not guaranteed to occur, for example, if GCD decides to map the global queue to a single CPU core.

broadwaylamb:
@glessard I'm curious though, how did you encounter this crash?

glessard:
There should not be a data race because `cancel()` is required to be thread-safe. This is in the contract for `Subscription`.
What got me down this road was trying to delay the `request` from a subscriber, but it may not have been this crash in particular. I found this one in particular by trying to reproduce/understand my original crash.

Issue #160, opened 2020-06-10
Question: Synthetize ObjectWillChange for @Published


nidegen:
Does OpenCombine manage automatic @Published linking to ObservableObject?

I tried to find any hints in the code, but there seem not to be any explicit code.

For Combine, they use something as described in this thread:
https://twitter.com/luka_bernardi/status/1155944329363349504?s=20

nidegen:
ah, found some talk on #97. Seems like an unsolved issue then?

nidegen:
Would there be a way to add `subscript<EnclosingSelf: ObservableObject>` only for ObservableObjects? A conditional extension on Published where EnclosingSelf conforms to ObservableObject? Guess that is impossible now

maximkrouk:
Maybe a custom property will work?

```swift
@propertyWrapper
struct OCombinePublished<Value> {
  public static subscript<EnclosingSelf: ObservableObject>(
    _enclosingInstance instance: EnclosingSelf,
    wrapped wrappedKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Value>,
    storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Self>
  ) -> Value where EnclosingSelf.ObjectWillChangePublisher {
    get { instance[keyPath: storageKeyPath].value }
    set {
      instance[keyPath: storageKeyPath].value = newValue
      instance.objectWillChange.send()
    }
  }
  
  var value: Value
  
  /// See ``Dependency`` discussion
  public init(wrappedValue: Value) {
    self.value = wrappedValue
  }
  
  public var wrappedValue: Value {
    get { value }
    set { value = newValue }
  }
}
```

PR #161, opened 2020-06-10
set ObservableObjectPublisher on Published

nidegen:
Solution to #160 and #97?

### DIFF
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -88,7 +88,7 @@ public struct Published<Value> {
         }
         set {
             if object[keyPath: storageKeyPath].objectWillChange == nil {
-                object[keyPath: storageKeyPath].objectWillChange = getObservablePublisher(object)
+                object[keyPath: storageKeyPath].objectWillChange = getPublisher(object)
             }
             object[keyPath: storageKeyPath].objectWillChange?.send()
             object[keyPath: storageKeyPath].publisher?.subject.send(newValue)
@@ -98,12 +98,14 @@ public struct Published<Value> {
     }
 }
 
-func getObservablePublisher<T:ObservableObject>(_ observable: T) -> ObservableObjectPublisher? {
-  return observable.objectWillChange as? ObservableObjectPublisher
+private func getPublisher<Observable:ObservableObject>(_ observable: Observable)
+                                                       -> ObservableObjectPublisher? {
+    return observable.objectWillChange as? ObservableObjectPublisher
 }
 
-func getObservablePublisher<T>(_ observable: T) -> ObservableObjectPublisher? {
-  return nil
+private func getPublisher<NonObservable>(_ observable: NonObservable) 
+                                         -> ObservableObjectPublisher? {
+    return nil
 }
 
 #else
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -87,6 +87,9 @@ public struct Published<Value> {
             return object[keyPath: storageKeyPath].value
         }
         set {
+            if object[keyPath: storageKeyPath].objectWillChange == nil {
+                object[keyPath: storageKeyPath].objectWillChange = getObservablePublisher(object)
+            }
             object[keyPath: storageKeyPath].objectWillChange?.send()
             object[keyPath: storageKeyPath].publisher?.subject.send(newValue)
             object[keyPath: storageKeyPath].value = newValue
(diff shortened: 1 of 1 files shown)


Sources/OpenCombine/Published.swift
```diff
@@ -87,13 +87,27 @@ public struct Published<Value> {
...
         // TODO: Benchmark and explore a possibility to use _modify
     }
 }
+
+private func getPublisher<Observable:ObservableObject>(_ observable: Observable)
+                                                       -> ObservableObjectPublisher? {
+    return observable.objectWillChange as? ObservableObjectPublisher
+}
+
+private func getPublisher<NonObservable>(_ observable: NonObservable)
```
OpenCombineBot, line 106: - :no_entry_sign: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/OpenCombine/Published.swift
```diff
@@ -87,13 +87,27 @@ public struct Published<Value> {
...
+                                         -> ObservableObjectPublisher? {
+    return nil
+}
```
broadwaylamb, line 109: I don't think this will work. Function overloads are resolved at compile time, not at run time. And because `object` in the setter is not constrained by `ObservableObject`, the first `getPublisher` will never ever be called.

nidegen: oh damn, true

nidegen: I guess there is really no way to make it work then?

broadwaylamb: Only something like #97 :(

nidegen:
https://github.com/apple/swift-evolution/blob/master/proposals/0267-where-on-contextually-generic.md will solve that, right?

broadwaylamb:
@nidegen I don't think so. This is still compile-time stuff. Apple uses runtime introspection in their implementation for a reason.

nidegen:
I was thinking that maybe it would work to add the static subscript function in a conextually generic extension. In principle, everything should be known at compile time, or am I missing something?

broadwaylamb:
>In principle, everything should be known at compile time, or am I missing something?

No. `Published` can't possibly know at compile time anything about the object that it's contained in.

nidegen:
OK. But [`EnclosingSelf`](https://forums.swift.org/t/property-wrappers-access-to-both-enclosing-self-and-wrapper-instance/32526) is not type-checked then? I am confused, sorry for the questions:)

broadwaylamb:
> OK. But [`EnclosingSelf`](https://forums.swift.org/t/property-wrappers-access-to-both-enclosing-self-and-wrapper-instance/32526) is not type-checked then?

Well, it is, but that doesn't help us. Even if we constrain `EnclosingSelf` to be `ObservableObject`, this won't help us, because things are the other way around really: it's not `Published` that sets its `objectWillChange` property by taking a publisher instance from `ObservableObject`'s `objectWillChange`. It's `ObservableObject` that should store the newly created publisher instance somewhere. And that "somewhere" is the `@Published` properties of the class that conforms to `ObservableObject`.

Simply put, we want the following semantics: suppose we have a class that conforms to `ObservableObject` and has some `@Published` properties. We wan't to be able to access the `objectWillChange` property once and get a publisher instance, then access it again and get _the same_ publisher instance. We can only do this by using some magic in the implementation of `objectWillChange`.

>I am confused, sorry for the questions:)

It's OK, I'm happy to answer them!

nidegen:
I see, getting the publisher from `ObservableObject` is impossible, I agree. But how are you so certain that it is not `Published` that takes the publisher from `ObservableObject`? What is bad about that?

If there is no way to do it the way you want it in normal Swift, I don't see why Apple would change so much to their version.

broadwaylamb:
>But how are you so certain that it is not Published that takes the publisher from ObservableObject? What is bad about that?

Because in order for the static subscript of `@Published` to be called, we need to access one of the `@Published` properties. But the `objectWillChange` publisher should be stored somewhere even if we never access any of the `@Published` properties in our client code.

In other words, this code should work:

```swift
class Foo: ObservableObject {
    @Published var i = 0

    init() {}
}

let foo = Foo()
assert(foo.observableObject === foo.observableObject)
```

Another thing is that we can't afford to access `ObservableObject`'s `objectWillChange` property in `Published` because the client can always implement `objectWillChange` themselves instead of relying on the default implementation. In that case, we'll have the situation where in OpenCombine that implementation is invoked when we access the static subscript in `Published`, but in Combine it's not invoked. This is the kind of semantic incompatibility that we want to avoid.

nidegen:
OK, I thought the subscript got called on init. But with the second argument I have to agree.

Looking at the [subscript proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#referencing-the-enclosing-self-in-a-wrapper-type), I still think they did it somehow like that:)

You mean:
```
assert(foo.observableObject === foo._i.observableObject)
```
right?

broadwaylamb:
>You mean:
>```
>assert(foo.observableObject === foo._i.observableObject)
>```
>right?

Nope. I've made a typo, it should be `foo.objectWillChange` instead of `foo.observableObject`, but otherwise what I've written is correct. Meaning that the `objectWillChange` property should not return a new publisher instance each time it is accessed.

nidegen:
ha, I did actually make the same mistake reading. Of course it is `objectWillChange`.

PR #163, opened 2020-06-12
Prepare for 0.9.0

### LINKED ISSUE #162: 0.9.0 release
It looks like enough good changes have accumulated since 0.8.0, @broadwaylamb would you mind tagging a new release? Or if this process is automated, would you mind sharing it, and I can do it myself if I may?

broadwaylamb:


README.md
```diff
@@ -69,6 +69,28 @@ Or enable the `-DOPENCOMBINE_COMPATIBILITY_TEST` compiler flag in Xcode's build
 
 > NOTE: Before starting to work on some feature, please consult the [GitHub project](https://github.com/broadwaylamb/OpenCombine/projects/2) to make sure that nobody's already making progress on the same feature! If not, then please create a draft PR to indicate that you're beginning your work.
 
+#### Releasing a new version
+
+1. Create a new branch from master and call it `release/<major>.<minor>.<patch>`.
+1. Replace the usages of the old version in `README.md` with the new version (make sure to check the [Swift Package Manager](#swift-package-manager) and [CocoaPods](#cocoapods) sections).
+1. Bump the version in `OpenCombine.podspec`, `OpenCombineDispatch.podspec` and `OpenCombineFoundation.podspec`. In the latter two you will also need to set the `spec.dependency "OpenCombine"` property to the **previous** version.
```
MaxDesiatov: >need to set the `spec.dependency "OpenCombine"` property to the **previous** version

Why previous version here?

broadwaylamb: Because otherwise the `pod lib lint` command that we run on our regular CI will fail when validating the `OpenCombineDispatch` and `OpenCombineFoundation` podspecs, since the dependencies are not yet in the trunk.

If we set the dependencies to the previous version (which is already in the trunk), everything will be fine.

This is purely to make the CI work. The clients will not experience any issues, since the version is specified with `>=`.

MaxDesiatov: Could you add the explanation to `README.md` then please? I imagine I'm not going to be the only one asking this "why" question 🙂

broadwaylamb: Yep

MaxDesiatov (approved):

PR #165, opened 2020-06-14
Implement OperationQueue scheduler

broadwaylamb:


Sources/OpenCombineFoundation/OperationQueue+Scheduler.swift
```diff
@@ -0,0 +1,287 @@
...
+
+                public init(floatLiteral value: TimeInterval) {
+                    magnitude = value
+                }
+
+                public init(_ timeInterval: TimeInterval) {
+                    magnitude = timeInterval
+                }
+
+                public init?<T>(exactly source: T) where T: BinaryInteger {
```
OpenCombineBot, line 101: - :no_entry_sign: Generic type name should be between 3 and 20 characters long: 'T' (`generic_type_name`)

PR #170, opened 2020-06-25
Rewrite PassthroughSubject, CurrentValueSubject and Future

broadwaylamb:
This PR introduces a more correct and efficient implementation for these.

No more leaking subscriptions!

### DIFF
--- a/Sources/OpenCombine/Helpers/ConduitBase.swift
+++ b/Sources/OpenCombine/Helpers/ConduitBase.swift
@@ -1,0 +1,40 @@
+//
+//  ConduitBase.swift
+//  
+//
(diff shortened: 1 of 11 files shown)


Sources/OpenCombine/Helpers/ConduitBase.swift
```diff
@@ -0,0 +1,40 @@
...
+        abstractMethod()
+    }
+
+    internal func cancel() {
+        abstractMethod()
+    }
+}
+
+extension ConduitBase: Equatable {
+    static func == (lhs: ConduitBase<Output, Failure>,
```
OpenCombineBot, line 30: - :no_entry_sign: All declarations should specify Access Control Level keywords explicitly. (`explicit_acl`)

PR #173, opened 2020-07-17
Fix a build error on linux: `kCFStringEncodingUTF8` is not defined.

adamleonard:
`kCFStringEncodingUTF8` is available in C, but not in swift. I think `CFStringBuiltInEncodings.UTF8.rawValue` is what's intended. However, I didn't take a close look at this code, so please let me know if I'm missing something.

Also fix a type error I was getting in a unit test with Swift 5.3.

### DIFF
--- a/Sources/OpenCombineFoundation/Timer+Publisher.swift
+++ b/Sources/OpenCombineFoundation/Timer+Publisher.swift
@@ -337,11 +337,19 @@ extension RunLoop.Mode {
         return CFRunLoopMode(rawValue as CFString)
 #else
         return rawValue.withCString {
-            CFStringCreateWithCString(
+#if swift(>=5.3)
+            return CFStringCreateWithCString(
                 nil,
                 $0,
                 CFStringBuiltInEncodings.UTF8.rawValue
             )
+#else
+            return CFStringCreateWithCString(
+                nil,
+                $0,
+                CFStringEncoding(kCFStringEncodingUTF8)
+            )
+#endif
         }
 #endif
     }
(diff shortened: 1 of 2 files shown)


broadwaylamb:
It seems like this is a breaking change in the core libraries that come with Swift 5.3. Could you please guard your change with `#if swift(>=5.3)` so that it would still work with earlier versions of Swift?

adamleonard:
Thanks for the super quick response. You're right, this does seem to be a swift 5.3 change on linux. Oddly, the fix compiles for me on swift 5.2 with macOS CoreFoundation, but not the linux CoreFoundation.

Pushed a commit that makes this conditional for 5.3. I verified the project compiles against the `swift:5.2-bionic` and `swiftlang/swift:nightly-5.3-bionic` docker images. Also, it looks like the iOS 9.3 sim used by the iOS circleci tests isn't available to download anymore?

Thanks for this amazing project! It's been a joy to work with.

Sources/OpenCombineFoundation/Timer+Publisher.swift
```diff
@@ -337,11 +337,19 @@ extension RunLoop.Mode {
...
+                $0,
+                CFStringBuiltInEncodings.UTF8.rawValue
+            )
+#else
+            return CFStringCreateWithCString(
                 nil,
                 $0,
                 CFStringEncoding(kCFStringEncodingUTF8)
             )
+#endif
```
broadwaylamb: Let's extract the encoding into a local variable so there's less code duplication:

```swift
#if swift(>=5.3)
let encoding = CFStringBuiltInEncodings.UTF8.rawValue
#else
let encoding = CFStringEncoding(kCFStringEncodingUTF8)
#endif // swift(>=5.3)

return CFStringCreateWithCString(
    nil,
    $0,
    encoding
)
```

broadwaylamb (approved):
Thank you!

PR #180, opened 2020-08-18
Add Carthage support

AYastrebov:
added carthage support


broadwaylamb:
To be honest, I'm not very comfortable with the idea of having an Xcode project checked in into source control, as it brings additional maintenance burden, like having to remember to synchronize the Xcode project with the file structure.

Apologies for the late response.

I will close it now.

PR #177, opened 2020-07-29
Incorrect argument label in call (have 'for:', expected 'forKey:')

devmaximilian:
This pull request relates to #176.

PR #176, opened 2020-07-29
Use of undeclared type PropertyListEncoder on Swift 5.0

devmaximilian:
`PropertyListEncoder` and `PropertyListDecoder` are both unavailable prior to Swift 5.1, causing a build error for Swift 5.0.

From what I've gathered they're available on the [swift-5.1-branch](https://github.com/apple/swift-corelibs-foundation/blob/swift-5.1-branch/Foundation/PropertyListEncoder.swift), but missing on the [swift-5.0-branch](https://github.com/apple/swift-corelibs-foundation/blob/swift-5.0-branch/Foundation/PropertyListEncoder.swift).

PR #185, opened 2020-10-01
Prepare for 0.10.1

devmaximilian:
Bumped version to include fix for build errors (#176 and #177) on linux when using swift 5.0 (swift-corelibs-foundation).


broadwaylamb (commented):
Thanks!

Could you please also update the version in the `README.md`?

MaxDesiatov (approved):
Thanks!

PR #197, opened 2020-12-01
Add OpenCombineShim product for easier importing

MaxDesiatov:
In cross-platform projects that need to support both Combine and OpenCombine it's a common pattern to use conditional imports:

```swift
#if canImport(Combine)
import Combine
#else
import OpenCombine
#endif
```

Adding these conditions all over the code is daunting. The proposed solution is a `CombineShim` module exposed as a product by OpenCombine, which re-exports Combine where available, and re-exports OpenCombine modules otherwise. It looks like this:

```swift
// The only content of the `CombineShim` module
#if canImport(Combine)
@_exported import Combine
#else
@_exported import OpenCombine
#if canImport(OpenCombineDispatch)
@_exported import OpenCombineDispatch
#endif
#if canImport(OpenCombineFoundation)
@_exported import OpenCombineFoundation
#endif
#endif
```
Thanks to this, such cross-platform projects can have their imports simplified to:

```swift
import CombineShim
```


Sources/CombineShim/CombineShim.swift
```diff
@@ -0,0 +1,11 @@
...
+@_exported import Combine
+#else
+@_exported import OpenCombine
+#if canImport(OpenCombineDispatch)
+@_exported import OpenCombineDispatch
+#endif
+#if canImport(OpenCombineFoundation)
+@_exported import OpenCombineFoundation
+#endif
+#endif
```
OpenCombineBot, line 11: - :no_entry_sign: Files should have a single trailing newline. (`trailing_newline`)

MaxDesiatov:
@broadwaylamb would you have a moment to take a look at this?

MaxDesiatov:
@broadwaylamb courteous ping

broadwaylamb:
LGTM.
I think it should be called `OpenCombineShim` for better namespacing. Or just something that starts with `OpenCombine`.

MaxDesiatov:
Good point, that's done now.

broadwaylamb (approved):

PR #201, opened 2020-12-22
Implementation for ObservableObject with Mirror

kateinoigakukun:
Ported from @MaxDesiatov and @broadwaylamb's work and stripped [wickwirew/Runtime](https://github.com/wickwirew/Runtime) dependency and unsafe operations.
This implementation only depends on `Mirror`.

### DIFF
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -107,8 +107,15 @@ public struct Published<Value> {
         case value(Value)
         case publisher(Publisher)
     }
+    @propertyWrapper
+    private class Box {
+        var wrappedValue: Storage
+        init(wrappedValue: Storage) {
+            self.wrappedValue = wrappedValue
+        }
+    }
 
-    private var storage: Storage
+    @Box private var storage: Storage
 
     internal var objectWillChange: ObservableObjectPublisher? {
         get {
@@ -119,7 +126,7 @@ public struct Published<Value> {
                 return publisher.subject.objectWillChange
             }
         }
-        set {
+        nonmutating set {
             projectedValue.subject.objectWillChange = newValue
         }
     }
@@ -145,14 +152,14 @@ public struct Published<Value> {
     ///
     /// - Parameter initialValue: The publisher's initial value.
     public init(wrappedValue: Value) {
-        storage = .value(wrappedValue)
+        _storage = Box(wrappedValue: .value(wrappedValue))
     }
 
     /// The property for which this instance exposes a publisher.
     ///
     /// The `projectedValue` is the property accessed with the `$` operator.
     public var projectedValue: Publisher {
-        mutating get {
+        get {
             switch storage {
             case .value(let value):
                 let publisher = Publisher(value)
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -110,6 +110,7 @@ public struct Published<Value> {
     @propertyWrapper
     private class Box {
         var wrappedValue: Storage
+
         init(wrappedValue: Storage) {
             self.wrappedValue = wrappedValue
         }
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -108,7 +108,7 @@ public struct Published<Value> {
         case publisher(Publisher)
     }
     @propertyWrapper
-    private class Box {
+    private final class Box {
         var wrappedValue: Storage
 
         init(wrappedValue: Storage) {
--- a/Sources/OpenCombine/Published.swift
+++ b/Sources/OpenCombine/Published.swift
@@ -128,7 +128,7 @@ public struct Published<Value> {
             }
         }
         nonmutating set {
-            projectedValue.subject.objectWillChange = newValue
+            getPublisher().subject.objectWillChange = newValue
         }
     }
 
@@ -160,15 +160,8 @@ public struct Published<Value> {
     ///
     /// The `projectedValue` is the property accessed with the `$` operator.
     public var projectedValue: Publisher {
-        get {
-            switch storage {
-            case .value(let value):
-                let publisher = Publisher(value)
-                storage = .publisher(publisher)
-                return publisher
-            case .publisher(let publisher):
-                return publisher
-            }
+        mutating get {
+            return getPublisher()
         }
         set { // swiftlint:disable:this unused_setter_value
             switch storage {
@@ -180,6 +173,17 @@ public struct Published<Value> {
         }
     }
 
+    /// Note: This method can mutate `storage`
+    internal func getPublisher() -> Publisher {
+        switch storage {
+        case .value(let value):
+            let publisher = Publisher(value)
+            storage = .publisher(publisher)
+            return publisher
+        case .publisher(let publisher):
+            return publisher
+        }
+    }
     // swiftlint:disable let_var_whitespace
     @available(*, unavailable, message: """
                @Published is only available on properties of classes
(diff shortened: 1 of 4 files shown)


Sources/OpenCombine/Published.swift
```diff
@@ -107,8 +107,15 @@ public struct Published<Value> {
         case value(Value)
         case publisher(Publisher)
     }
+    @propertyWrapper
+    private class Box {
+        var wrappedValue: Storage
+        init(wrappedValue: Storage) {
```
OpenCombineBot, line 113: - :no_entry_sign: Let and var should be separated from other statements by a blank line. (`let_var_whitespace`)

MaxDesiatov (changes requested):
There are a few more tests [in the original PR](https://github.com/OpenCombine/OpenCombine/pull/97/files#diff-17b0d56684749e6aad111141e0945b7b13ca6829f85120990cf8f759cf6dc54e) that could be added here as well.

MaxDesiatov (approved):
👏

MaxDesiatov:
Hi @broadwaylamb, would you have a moment to look at this? The lack of working `ObservableObject` is the only major difference between this repo and our SwiftWasm fork, would be amazing to work with you on getting this reviewed and merge if possible.

Sources/OpenCombine/Published.swift
```diff
@@ -107,8 +107,16 @@ public struct Published<Value> {
         case value(Value)
         case publisher(Publisher)
     }
+    @propertyWrapper
+    private class Box {
```
broadwaylamb:
```suggestion
    private final class Box {
```

Sources/OpenCombine/Published.swift
```diff
@@ -145,14 +153,14 @@ public struct Published<Value> {
...
-        storage = .value(wrappedValue)
+        _storage = Box(wrappedValue: .value(wrappedValue))
     }
 
     /// The property for which this instance exposes a publisher.
     ///
     /// The `projectedValue` is the property accessed with the `$` operator.
     public var projectedValue: Publisher {
-        mutating get {
+        get {
```
broadwaylamb: In Apple's Combine, this getter is `mutating`, even though it doesn't actually mutate anything. Let's be compatible.

Tests/OpenCombineTests/ObservableObjectTests.swift
```diff
@@ -0,0 +1,423 @@
...
+
+    func testNoFields() {
+        let observableObject = NoFields()
+        _ = observableObject.objectWillChange
+    }
+
+    func testNoPublishedFields() {
+        let observableObject = NoPublishedFields()
+        _ = observableObject.objectWillChange
+    }
```
broadwaylamb: Please add a TODO comment to these tests, so that in the future, when we improve the implementation, we could verify that the returned `ObservableObjectPublisher` is the same every time

Tests/OpenCombineTests/ObservableObjectTests.swift
```diff
@@ -0,0 +1,423 @@
...
+        XCTAssertEqual(observableObject.value2, false)
+    }
+
+    func testGenericSubclassOfResilientClass() {
+        let observableObject = ResilientClassGenericSubclass("hello", true)
+
+        var counter = 0
+
+        // A bug in Combine (FB7471594). It should not crash. Why would it crash?
+        assertCrashesOnDarwin {
```
broadwaylamb: This bug has been fixed, it doesn't crash anymore. Let's remove this.

Tests/OpenCombineTests/ObservableObjectTests.swift
```diff
@@ -0,0 +1,423 @@
...
+        }
+    }
+
+    func testGenericSubclassOfResilientClass2() {
+        let observableObject = ResilientClassGenericSubclass2("hello", true)
+
+        var counter = 0
+
+        // A bug in Combine (FB7471594). It should not crash. Why would it crash?
+        assertCrashesOnDarwin {
```
broadwaylamb: And this as well.

Tests/OpenCombineTests/Helpers/AssertCrashes.swift
```diff
@@ -84,4 +84,13 @@ extension XCTest {
...
     }
+
+    @available(macOS 10.13, iOS 8.0, *)
+    func assertCrashesOnDarwin(within body: () -> Void) {
+#if canImport(Darwin) && OPENCOMBINE_COMPATIBILITY_TEST
+        assertCrashes(within: body)
+#else
+        body()
+#endif
+    }
```
broadwaylamb: Since this function is no longer used, let's remove it.

broadwaylamb (commented):
Sorry for the delay. Left some comments, overall this looks very good as a temporary solution, until I come up with a proper implementation which will also work with classes that conform to `CustomReflectable`.

Sources/OpenCombine/Published.swift
MaxDesiatov: This breaks the build as it requires us to remove `nonmutating` from `set` on `internal var objectWillChange`:

```
 Published.swift:131

   NOTE   cannot use mutating getter on immutable value: 'self' is immutable
  131 |             projectedValue.subject.objectWillChange = newValue
      |              ^
```

@kateinoigakukun WDYT? Are there other options to work around the error?

Tests/OpenCombineTests/ObservableObjectTests.swift
kateinoigakukun: Hmm, it seems this bug is not fixed on macOS 10.15.7. Which version are you using?

Tests/OpenCombineTests/ObservableObjectTests.swift
kateinoigakukun: ditto

Tests/OpenCombineTests/ObservableObjectTests.swift
broadwaylamb: The latest iOS. Combine on macOS is always a little bit behind its iOS version.
I case of Catalina, it is _quite_ behind.

kateinoigakukun: So should we run this test suit on only iOS?

broadwaylamb: When testing against Combine — yes. But this applies to all the tests, not just this.

kateinoigakukun: I addressed 👍

Tests/OpenCombineTests/ObservableObjectTests.swift
```diff
@@ -0,0 +1,432 @@
...
+
+        observableObject.value2.toggle()
+
+        XCTAssertEqual(counter, 2)
+        XCTAssertEqual(observableObject.value2, false)
+    }
+
+    // A bug in Combine (FB7471594). This has been fixed on iOS.
+    // But not deployed on other Darwin OS yet
+    #if !OPENCOMBINE_COMPATIBILITY_TEST || os(iOS)
```
broadwaylamb: No, this condition is not needed. What I mean is that one should make sure to only run compatibility tests _against the most recent version_ of Combine (since that's what we aim to be compatible with). It so happens that the most recent version of Combine is usually vendored in the newest iOS version (not even the newest macOS version, and certainly not Catalina).

Tests should not be disabled based on that.

Just FYI, you can run the compatibility tests in the iOS simulator, this way they should pass. This is exactly what we do on our CI build.

kateinoigakukun: done

broadwaylamb (approved):
Thank you!

PR #209, opened 2021-07-05
support library evolution

bestswifter:
These changes are made according to the `.swiftinterface` of Apples's Official Combine framework.

So it should not affect the performance or correctness of OpenCombine.

With these changes, OpenCombine can be distributed as a static binary framework

### DIFF
--- a/Sources/OpenCombine/ImmediateScheduler.swift
+++ b/Sources/OpenCombine/ImmediateScheduler.swift
@@ -54,7 +54,6 @@ public struct ImmediateScheduler: Scheduler {
             public var magnitude: Int
 
             /// Creates an immediate scheduler time interval from the given time interval.
-            @inlinable
             public init(_ value: Int) {
                 magnitude = value
             }
--- a/Sources/OpenCombine/AnyPublisher.swift
+++ b/Sources/OpenCombine/AnyPublisher.swift
@@ -62,6 +62,7 @@ extension Publisher {
 ///
 /// You can use OpenCombine’s `eraseToAnyPublisher()` operator to wrap a publisher with
 /// `AnyPublisher`.
+@frozen
 public struct AnyPublisher<Output, Failure: Error>
   : CustomStringConvertible,
     CustomPlaygroundDisplayConvertible
@@ -112,7 +113,7 @@ extension AnyPublisher: Publisher {
 
 /// A type-erasing base class. Its concrete subclass is generic over the underlying
 /// publisher.
-@usableFromInline
+@usableFromInline @_fixed_layout
 internal class PublisherBoxBase<Output, Failure: Error>: Publisher {
 
     @inlinable
@@ -126,7 +127,7 @@ internal class PublisherBoxBase<Output, Failure: Error>: Publisher {
     }
 }
 
-@usableFromInline
+@usableFromInline @_fixed_layout
 internal final class PublisherBox<PublisherType: Publisher>
     : PublisherBoxBase<PublisherType.Output, PublisherType.Failure>
 {
--- a/Sources/OpenCombine/AnySubscriber.swift
+++ b/Sources/OpenCombine/AnySubscriber.swift
@@ -11,6 +11,7 @@
 /// expose. You can also use `AnySubscriber` to create a custom subscriber by providing
 /// closures for the methods defined in `Subscriber`, rather than implementing
 /// `Subscriber` directly.
+@frozen
 public struct AnySubscriber<Input, Failure: Error>: Subscriber,
                                                     CustomStringConvertible,
                                                     CustomReflectable,
@@ -130,7 +131,7 @@ public struct AnySubscriber<Input, Failure: Error>: Subscriber,
 
 /// A type-erasing base class. Its concrete subclass is generic over the underlying
 /// subscriber.
-@usableFromInline
+@usableFromInline @_fixed_layout
 internal class AnySubscriberBase<Input, Failure: Error>: Subscriber {
 
     @inline(__always)
@@ -157,7 +158,7 @@ internal class AnySubscriberBase<Input, Failure: Error>: Subscriber {
     }
 }
 
-@usableFromInline
+@usableFromInline @_fixed_layout
 internal final class AnySubscriberBox<Base: Subscriber>
     : AnySubscriberBase<Base.Input, Base.Failure>
 {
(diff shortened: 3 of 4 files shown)


Sources/OpenCombine/ImmediateScheduler.swift
```diff
@@ -54,7 +54,6 @@ public struct ImmediateScheduler: Scheduler {
             public var magnitude: Int
 
             /// Creates an immediate scheduler time interval from the given time interval.
-            @inlinable
             public init(_ value: Int) {
                 magnitude = value
             }
```
broadwaylamb, line 59: Why remove it?

broadwaylamb (commented):
Thanks for the PR. However, I'm not sure why someone would want OpenCombine as a binary framework, considering that it is open source and there is CocoaPods and SPM support. Do you have a specific use case?

Also, I really want to use as little underscored attributes (like `@_fixed_layout`) as possible, due to them being unstable.

Sources/OpenCombine/ImmediateScheduler.swift
bestswifter: OpenCombine is a foundational framework and can be depended by many other frameworks. Actually I commit this PR because I want to distribute my framework as binary. So all of my dependencies should be compiled when turning on `BUILD_LIBRARY_FOR_DISTRIBUTION`.

As of the underscored attributes, it appears in Combines's interface, I think maybe we can keep pace with Apple? After all, they are distributing Combine as a dynamic linking framework.

The `@inlinable` attribute is removed because it cannot be compiled when tuning on `BUILD_LIBRARY_FOR_DISTRIBUTION` because this `init` method maybe inlined into client code. Basically, there are two approaches to solve this problem:

1. Mark the struct/class as `@frozen` or `@_fixed_layout` so that compiler now knows the structure will not be changed in the future, and the memory layout is fixed instead of flexible
2. Remove the `@inlinable` attribute to disable cross-module inline optimization

Here I chose the second solution because `Stride` is neither `@frozen` or `@_fixed_layout`, and the `init` mehod does not appear in `.swiftinterface` of Combine. If it is `@inlinable` in Apple's implementation, it will be exported as part of the module and shown in the `.swiftinterface` file.

bestswifter: Use OpenCombine as a binary framework can also save some compiling time for almost every building for every app.

SPM now also supports distributing packages as a binary framework now. See: https://developer.apple.com/documentation/swift_packages/distributing_binary_frameworks_as_swift_packages

bestswifter: I have no idea why test failed on Ubuntu 18.04 (Swift 5.0) since `@frozen` was supported by Swift 5.0

The other two failure may disappear by retrying the test.

broadwaylamb: >I have no idea why test failed on Ubuntu 18.04 (Swift 5.0) since `@frozen` was supported by Swift 5.0

It seems it wasn't.

>As of the underscored attributes, it appears in Combines's interface, I think maybe we can keep pace with Apple? After all, they are distributing Combine as a dynamic linking framework.

They do it only because they don't want to distribute Combine's source code. For libraries whose source code they _do_ distribute (for example, [swift-algorithms](https://github.com/apple/swift-algorithms)), they don't support library evolution.

>Use OpenCombine as a binary framework can also save some compiling time for almost every building for every app.

At the cost of some runtime performance (probably negligible though).

___

I'm not sure how I feel about this. I hear your point and see why it is important for you, but if we start supporting binary frameworks, we inevitably will have to think about binary stability each time we make a change to the library, which is something I would really like to avoid.

Does your framework uses OpenCombine types in its public API?
- If yes, I wonder what happens if a client of your framework also depends on OpenCombine, but built in non-library evolution mode, or of different version of it.
- If no, you could probably get away with using OpenCombine as a private dependency by using `@_implementationOnly import OpenCombine`. This way you don't need OpenCombine to be built in library evolution mode.

If the only acceptable solution for you is to build OpenCombine in library evolution mode nonetheless, is it possible for you to get away without adding all those `@frozen` and `@_fixed_layout` attributes?

bestswifter: > If yes, I wonder what happens if a client of your framework also depends on OpenCombine, but built in non-library evolution mode, or of different version of it.

YES. Both my framework and clients are now distributed as source code. If my framework is built and distributed in non-library evolution mode, my client has to rebuild all frameworks depending on mine every time I release a new version because my module is not stable. But my framework is also an underlying foundational framework, which is depended by many other ones and I don't want to show my source code.

> we inevitably will have to think about binary stability each time we make a change to the library, which is something I would really like to avoid.

Totally Agree!!! And I'm facing the same problem. However, hiding source code and avoid compiling are two common and very important features for those frameworks which depend on OpenCombine. What I think very lucky of OpenCombine is, Apple has concerned about the trade off between the flexibility and performance ahead of us. For whatever reason, they have stated that  the memory layout of some struct and class will not be changed by marking them as `@froze` or `@_fixed_layout `

> is it possible for you to get away without adding all those @frozen and @_fixed_layout attributes?

Again, we can't keep flexible without losing performance.

In conclusion, just from my opinion, the loss of flexibility seems ok because Apple did the same thing and has proved the feasibility. There is only one occurrence of runtime performance loss. It will add an extra function call to set an `Int` value which is acceptable me. And there is no doubt that you have the power to make decision.

broadwaylamb: > > If yes, I wonder what happens if a client of your framework also depends on OpenCombine, but built in non-library evolution mode, or of different version of it.
> 
> YES. Both my framework and clients are now distributed as source code. If my framework is built and distributed in non-library evolution mode, my client has to rebuild all frameworks depending on mine every time I release a new version because my module is not stable. But my framework is also an underlying foundational framework, which is depended by many other ones and I don't want to show my source code.

I mean, there could be a version conflict if you and you clients use different versions of OpenCombine.

What do you think about the `@_implementationOnly` import solution?

I just really don't want to increase the maintenance burden unless the ability to build OpenCombine in library evolution mode is desired by a significant number of OpenCombine users. I hope you understand. I'm going to close this PR for now, but I invite you to create an issue in this repo, and if more people give it a thumbs up, I'll consider reopening it.

CrazyFanFan:
I also need library evolution, and here is a brief background: I work on a large-scale Swift project, where we use binary dependencies during development to reduce build time, although we still use source code builds when releasing.  Since modules with BUILD_LIBRARY_FOR_DISTRIBUTION enabled require dependencies that also have BUILD_LIBRARY_FOR_DISTRIBUTION enabled, I had to fork a copy of the OpenCombine code and make some modifications to support library evolution.  However, I have always hoped that we could return to the main branch, i.e., using the public release version.

PR #210, opened 2021-07-08
Fix reentrancy bugs in subscribers

### LINKED ISSUE #208: Should change unfair lock to unfair recursive lock in Sink?
when AnyCancellable calls cancel function automatically in DEINIT, Sink's cancel function will be called, then terminateAndConsumeLock will be called. As the result, the receiveValue block and receiveCompletion block will be set to nil. Then objects in receiveValue block or receiveCompletion block may be destroyed. If there is an object that in block, it will call same sink's receive completion to finish in object's DEINIT function. That will make a crash cause call lock's lock function twice without unlock.

broadwaylamb:
Fixes #208.

### DIFF
--- a/Tests/OpenCombineTests/Helpers/AutomaticallyFinish.swift
+++ b/Tests/OpenCombineTests/Helpers/AutomaticallyFinish.swift
@@ -36,6 +36,14 @@ final class AutomaticallyFinish<Output, Failure: Error> {
     }
 }
 
+extension AutomaticallyFinish: Publisher {
+    func receive<Downstream: Subscriber>(subscriber: Downstream)
+        where Downstream.Failure == Failure, Downstream.Input == Output
+    {
+        publisher.subscribe(subscriber)
+    }
+}
+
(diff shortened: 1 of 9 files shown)


Tests/OpenCombineTests/Helpers/AutomaticallyFinish.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    func notify(_ value: Output) {
+        _ = publisher.send(value)
+    }
+
+    func listen(receiveCompletion: @escaping (Subscribers.Completion<Failure>) -> Void,
+                receiveValue: @escaping (Output) -> Void) -> AnyCancellable {
+        return publisher.sink(receiveCompletion: receiveCompletion,
+                            receiveValue: receiveValue)
```
OpenCombineBot, line 35: - :no_entry_sign: Function parameters should be aligned vertically if they're in multiple lines in a method call. (`vertical_parameter_alignment_on_call`)

PR #216, opened 2021-08-02
Add support for Windows

MortenGregersen:
It looks like the bug mentioned in `Package.swift` is fixed in Xcode 12.2 or later (https://bugs.swift.org/browse/SR-13817). Shouldn't we support Windows then?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,8 +11,7 @@ let supportedPlatforms: [Platform] = [
     .tvOS,
     .linux,
     .android,
-    // Disable Windows because of https://bugs.swift.org/browse/SR-13817
-    // .windows,
+    .windows,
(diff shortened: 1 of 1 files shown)


broadwaylamb (changes requested):
Thank you!

As I see, this change broke the build for older Xcode versions.
Let's add a compilation condition to only enable it since Xcode 12.2 (which corresponds to Swift 5.3.1):

```swift
#if compiler(>=5.3.1)
// Conditionally support Windows because of https://bugs.swift.org/browse/SR-13817
supportedPlatforms.append(.windows)
#endif
```

MaxDesiatov:
Closing as superseded by #220

PR #219, opened 2021-08-27
Update for Xcode 13

### LINKED ISSUE #227: [Feature Request] Async Publisher.values
Combine added a [`values` property](https://developer.apple.com/documentation/combine/publisher/values-1dm9r) (and a [throwing counterpart](https://developer.apple.com/documentation/combine/publisher/values-v7nz)) that lets you bridge from the world of Combine into async/await (on supported Swift version), and we don't have parity here.

Here's my attempt to implement this:

```swift
#if !canImport(Combine)

import OpenCombine

// TODO: write throwing variant, as well.
extension Publisher where Failure == Never {
    var values: AsyncPublisher<Self> {
        return AsyncPublisher(self)
    }
}

// Trying to keep the type names the same as Combine
struct AsyncPublisher<P: Publisher>: AsyncSequence where P.Failure == Never {
    typealias Element = P.Output

    let p: P

    init(_ p: P) {
        self.p = p
    }

    func makeAsyncIterator() -> Iterator { Iterator(p) }

    struct Iterator: AsyncIteratorProtocol {
        // Does this cancellable ever need to be nil-ed out? IDK.
        private let cancellable: AnyCancellable?
        private var streamIterator: AsyncStream<P.Output>.Iterator

        init(_ p: P) {
            var cancellable: AnyCancellable? = nil

            let stream = AsyncStream(P.Output.self) { continuation in 
                cancellable = p
                    .handleEvents(receiveCancel: { continuation.finish() })
                    .sink(
                        receiveCompletion: { _ in continuation.finish() },
                        receiveValue: { continuation.yield($0) }
                    )
            }

            self.cancellable = cancellable
            self.streamIterator = stream.makeAsyncIterator()
        }

        public mutating func next() async -> P.Output? {
            return await streamIterator.next()
        }
    }
}
#endif
```

It's a little jank, so I don't feed confident opening a PR for it yet, but I'd appreciate some feedback and ideas!

broadwaylamb:
- [x] Update tests to pass in Combine compatibility mode
- [x] Fix OpenCombine so all tests pass
- [x] Implement the new APIs
- [x] Write tests for the new APIs
- [x] Enable testing with Swift 5.5 on Ubuntu (when Swift 5.5 docker image is released)


Sources/OpenCombineFoundation/NotificationCenter.swift
```diff
@@ -200,13 +200,11 @@ extension Notification {
 
         func cancel() {
             lock.lock()
-            guard let center = self.center, let observation = self.observation else {
+            guard let center = self.center.take(), let observation = self.observation.take() else {
```
OpenCombineBot, line 203: - :no_entry_sign: Line should be 90 characters or less: currently 99 characters (`line_length`)

MaxDesiatov:
@broadwaylamb would you mind if I merge this as is, without tests? I'm thinking of tagging a new version real soon and it would be great to have this included.

broadwaylamb:
@MaxDesiatov to be honest, I'm not sure this implementation works properly. Tests are most certainly needed.

MaxDesiatov:
ok, I'll postpone this then

.github/workflows/wasm.yml
```diff
@@ -7,10 +7,28 @@ on:
     branches: [master]
 
 jobs:
-  carton_wasmer_test:
+  carton_wasmer_test_5_3:
+    name: "Execute tests on WASM (Swift 5.3)"
```
MaxDesiatov: nit: [according to the WebAssembly spec](https://webassembly.github.io/spec/core/intro/introduction.html#id1), correct capitalization is Wasm
```suggestion
    name: "Execute tests on Wasm (Swift 5.3)"
```

.github/workflows/wasm.yml
```diff
@@ -7,10 +7,28 @@ on:
...
     runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
       - uses: swiftwasm/swiftwasm-action@v5.3
+  
+  carton_wasmer_test_5_4:
+    name: "Execute tests on WASM (Swift 5.4)"
```
MaxDesiatov:
```suggestion
    name: "Execute tests on Wasm (Swift 5.4)"
```

broadwaylamb:
@MaxDesiatov I'm trying to get async tests to work with Wasm, and I stumbled upon the following issue: apparently, the swiftwasm fork of XCTest doesn't have anything async-related — `XCTestExpectation`, `XCTWaiter`, etc.

Swiftwasm's Foundation doesn't have `RunLoop`, either.

So, is there any way for me to wait on a `Task` (the stdlib class) in a blocking manner, so that I could call an `async` closure inside a sync one and wait for it to finish execution?

MaxDesiatov:
Unfortunately, currently the answer is no, I think. AFAIU Swift 5.5.0 didn't have support for `async` tests. Some later 5.5.x release of upstream Swift apparently added support for async tests, but I didn't merge that into our fork yet. And that merge is unlikely to be trivial. IDK if it makes sense to do it in 5.5 point release of SwiftWasm at all, we may just do it for 5.6 only, since that one is imminent, and our resources are limited. 

I think for now it's best to exclude async tests with `#if !os(WASI)`.

@kateinoigakukun WDYT? Am I wrong in saying there's no easy way around this right now?

MaxDesiatov:
Also, be careful with Windows 5.5.2, there are some regressions there https://github.com/MaxDesiatov/swift-windows-action/issues/1

broadwaylamb:
sigh…

PR #220, opened 2021-09-08
Test with Windows on GitHub Actions

MaxDesiatov:


broadwaylamb:
Awesome, thank you!

I can see that the C helpers aren’t compiling because of missing `pthread.h`. I think we could use `std::mutex` from C++ whenever pthread is not available.

MaxDesiatov:
Would you prefer `std::mutex` to be always used on all platforms, or to defer to `pthread` on platforms only where the latter is available? I wonder if the former would be more convenient and reduce the amount if `#if` checks we have to do, as a more portable option.

broadwaylamb:
Always using `std::mutex` would be convenient, but there are two reasons I haven’t already went with it:
- on Apple platforms, `os_unfair_lock` is much much faster than `pthread_mutex_lock` (and `std::mutex`, which uses `pthread_mutex_lock` on Darwin and Linux)
- pthread API allows to set the `PTHREAD_MUTEX_ERRORCHECK` attribute for mutexes, which enables checking for accidental recursive captures of non-recursive mutexes. `std::mutex` doesn’t allow that, and recursively capturing it is UB.

broadwaylamb:
So, it now compiles, but we hit [this](https://github.com/apple/swift-tools-support-core/blob/492fe8dbf50c6f906444dbc227d3edb35ab0f4a7/Sources/TSCUtility/Triple.swift#L160).

broadwaylamb:
https://forums.swift.org/t/error-tscutility-triple-swift-fatal-error-could-not-determine-host-triple-malformed/47775/4

Seems like re-running the job helps.

MaxDesiatov:
`import CoreFoundation` won't work on Windows since CoreFoundation is now hidden as a private implementation detail on non-Darwin platforms

broadwaylamb:
Well that's pretty stupid.

MaxDesiatov:
TWIMC, more details available here https://forums.swift.org/t/formalizing-the-unavailability-of-core-foundation/40216

broadwaylamb:
Thanks!

Tests/OpenCombineTests/Helpers/ExecuteOnBackgroundThread.swift
```diff
@@ -52,36 +86,59 @@ func executeOnBackgroundThread<ResultType>(
...
+#else
+        handle.deallocate()
+#endif
+    }
+
+    func join() -> ResultType {
+        precondition(!joined, "Already joined!")
+        joined = true
+#if os(Windows)
+        WaitForSingleObject(handle, INFINITE);
```
OpenCombineBot, line 113: - :no_entry_sign: Lines should not have trailing semicolons. (`trailing_semicolon`)

Tests/OpenCombineTests/Helpers/ExecuteOnBackgroundThread.swift
```diff
@@ -52,36 +87,59 @@ func executeOnBackgroundThread<ResultType>(
...
```
OpenCombineBot, line 114: - :no_entry_sign: Lines should not have trailing semicolons. (`trailing_semicolon`)

broadwaylamb:
So, we now officially support Windows.

Congratulations and thank you!

broadwaylamb:
Let me know if it's safe to clean up the commit history — I don't want to force push to your branch without your approval.

MaxDesiatov:
Yeah, fine to clean up, no problem

broadwaylamb (approved):

PR #226, opened 2021-11-22
Migrate SwiftLint checks from CircleCI to GHA

MaxDesiatov:
SwiftLint integration with Danger no longer works on CircleCI. I'm migrating it to GitHub Actions in a way that's known to work in other repositories.

I've also added a job for testing with Swift 5.5 on Ubuntu 18.04.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -241,6 +241,7 @@ workflows:
       - "Execute tests on Ubuntu 18.04 (Swift 5.2)"
       - "Execute tests on Ubuntu 18.04 (Swift 5.3)"
       - "Execute tests on Ubuntu 18.04 (Swift 5.4)"
+      - "Execute tests on Ubuntu 18.04 (Swift 5.5)"
   "OpenCombine: validate podspec files":
     jobs:
       - "Run Pod spec lint"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -205,24 +205,12 @@ jobs:
       SWIFT_VERSION: "5.4"
     <<: *ubuntu_tests_steps
 
-  "Run SwiftLint and Danger":
-    macos:
-      xcode: "11.4.0"
+  "Execute tests on Ubuntu 18.04 (Swift 5.5)":
```
broadwaylamb, line 208: You forgot to add this job below to the `jobs` list.

broadwaylamb (changes requested):

.circleci/config.yml
MaxDesiatov: Great catch! These CircleCI configs are somewhat more confusing than GHA configs to me. I wonder if it's worth considering a migration from CircleCI to GHA for Linux jobs at least, at some point in the future.

broadwaylamb: I agree, I was thinking about this too.

PR #236, opened 2022-11-13
Add primary associated types

### LINKED ISSUE #237: Add primary associated type support
When Swift released 5.7 in Xcode 14.0, Combine framework has updated its protocol interface to support primary associated type support.

eg.
```swift
// Latest Combine
public protocol Publisher<Output, Failure>
```
```swift
// Current OpenCombine and old Combine(iOS 15-)
public protocol Publisher
```

Adding such feature support enable us to use `some Publisher<Int, Error>` without needing to convert it to AnyPublisher.

```swift
func loadUser(
    withID id: User.ID
) -> some Publisher<User, Error> {
    urlSession
        .dataTaskPublisher(for: urlForLoadingUser(withID: id))
        .map(\.data)
        .decode(type: User.self, decoder: decoder)
}
```

See more info at [Combining opaque return types with primary associated types](https://www.swiftbysundell.com/articles/opaque-return-types-primary-associated-types)

danny1113:
Apple's official Combine framework also implemented primary associated types feature introduced in swift 5.7.

https://developer.apple.com/documentation/combine/connectablepublisher
https://developer.apple.com/documentation/combine/publisher
https://developer.apple.com/documentation/combine/subscriber
https://developer.apple.com/documentation/combine/subject
https://developer.apple.com/documentation/combine/scheduler

so I changed `ConnectablePublisher`, `Publisher`, `Subscriber`, `Subject`, `Scheduler` to support primary associated types as well.

### DIFF
--- a/Sources/OpenCombine/ConnectablePublisher.swift
+++ b/Sources/OpenCombine/ConnectablePublisher.swift
@@ -5,6 +5,25 @@
 //  Created by Sergej Jaskiewicz on 14.06.2019.
 //
 
+#if swift(>=5.7)
+/// A publisher that provides an explicit means of connecting and canceling publication.
+///
+/// Use a `ConnectablePublisher` when you need to perform additional configuration or
+/// setup prior to producing any elements.
+///
+/// This publisher doesn’t produce any elements until you call its `connect()` method.
+///
+/// Use `makeConnectable()` to create a `ConnectablePublisher` from any publisher whose
+/// failure type is `Never`.
+public protocol ConnectablePublisher<Output, Failure>: Publisher {
+
+    /// Connects to the publisher, allowing it to produce elements, and returns
+    /// an instance with which to cancel publishing.
+    ///
+    /// - Returns: A `Cancellable` instance that you use to cancel publishing.
+    func connect() -> Cancellable
+}
+#else
 /// A publisher that provides an explicit means of connecting and canceling publication.
 ///
 /// Use a `ConnectablePublisher` when you need to perform additional configuration or
@@ -22,3 +41,4 @@ public protocol ConnectablePublisher: Publisher {
     /// - Returns: A `Cancellable` instance that you use to cancel publishing.
     func connect() -> Cancellable
 }
+#endif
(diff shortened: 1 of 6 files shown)


Kyle-Ye:
Should we add a Unit Test file for this feature in case somebody accidentally delete it?

For example:
```swift
import XCTest

#if swift(>=5.7)

#if OPENCOMBINE_COMPATIBILITY_TEST
import Combine
@available(macOS 10.15, iOS 13.0, *)
private typealias Published = Combine.Published
@available(macOS 10.15, iOS 13.0, *)
private typealias ObservableObject = Combine.ObservableObject
#else
import OpenCombine
private typealias Published = OpenCombine.Published
private typealias ObservableObject = OpenCombine.ObservableObject
#endif
@available(macOS 10.15, iOS 13.0, *)
final class PrimaryAssociatedTypeTests: XCTestCase {
    func testCombinePrimaryAssociatedTypes() {
        let p1: some Publisher<Int, Error> = Empty()
  		...
	}
}
```

danny1113:
Ok sure.

Kyle-Ye:
> I'll merge it as soon as I fix the failing compatibility tests.
> https://github.com/OpenCombine/OpenCombine/issues/237#issuecomment-1318935543

Sorry to bother you. Is there any progress on this PR? @broadwaylamb

NikolayJuly:
I also interested in this PR. But do we really need check for swift version?
First of all - there are tags, do we need support in older swift version? People just can ference older point in time, if they need stopped on swift 5.5 compiler, for example. 
Second, it will make simplerto maintain, with only one implementation.
Am I missing anything?
I think it worth just implement primary associated types and remove swift version checks.

danny1113:
Because swift-tools-version in Package.swift is 5.5, so I think maybe we should also support Swift 5.5 as well.
Or we could bump the swift-tools-version to 5.7 and remove those version checks.

NikolayJuly:
Tbh, I would prefer version bump in package. It will really simplify support. Less code is always better

broadwaylamb:
Support for primary associated types has been merged to master in #239. I've also published a new release. Thank you!

PR #244, opened 2023-10-11
COpenCOmbineHelpers: add an explicit modulemap

compnerd:
Introduce an explicit modulemap as not all build systems will write out a modulemap for you.  Add the explicit modulemap to allow building with alternative build systems.


tristanlabelle (approved):

broadwaylamb (approved):
Thank you!

compnerd:
@broadwaylamb not sure what else needs to be done here.

